PR #110, opened 2019-08-07
Issue #102 mergeAggrs and mergePayoutsIntoBalances

### LINKED ISSUE #102: Implement mergeAggrs & mergePayoutsIntoBalances
For reference see the [JS implementation](https://github.com/AdExNetwork/adex-validator/blob/4c9178501da5e8d4d0784d47b17d3fac2cf98063/services/validatorWorker/lib/mergeAggrs.js)

elpiel:
resolves #102

### DIFF
--- a/validator/src/domain/event.rs
+++ b/validator/src/domain/event.rs
@@ -7,8 +7,10 @@ use domain::{AdUnit, BigNum};
 #[serde(tag = "type", rename_all = "SCREAMING_SNAKE_CASE")]
 #[derive(Serialize, Deserialize)]
 pub enum Event {
+    #[serde(rename_all = "camelCase")]
     Impression {
         publisher: String,
+        ad_unit: AdUnit,
     },
     #[serde(rename_all = "camelCase")]
     UpdateImpressionPrice {
@@ -16,12 +18,15 @@ pub enum Event {
         price: u64,
         ad_unit: AdUnit,
     },
+    #[serde(rename_all = "camelCase")]
     ImpressionWithCommission {
         earner: Vec<Earner>,
+        ad_unit: AdUnit,
     },
-    ImpressionPricePerCase {},
+    #[serde(rename_all = "camelCase")]
     Pay {
         outputs: HashMap<String, BigNum>,
+        ad_unit: AdUnit,
     },
     PauseChannel {},
     #[serde(skip_deserializing)]
--- a/validator/src/domain/event.rs
+++ b/validator/src/domain/event.rs
@@ -1,11 +1,12 @@
 use std::collections::HashMap;
 
+use chrono::{DateTime, Utc};
+use num_traits::CheckedSub;
 use serde::{Deserialize, Serialize};
 
-use adapter::ChannelId;
-use chrono::{DateTime, Utc};
+use domain::balances_map::get_balances_after_fees_tree;
 use domain::validator::message::Accounting;
-use domain::{AdUnit, BalancesMap, BigNum, Channel, DomainError};
+use domain::{AdUnit, BalancesMap, BigNum, Channel, ChannelId, DomainError};
 
 #[serde(tag = "type", rename_all = "SCREAMING_SNAKE_CASE")]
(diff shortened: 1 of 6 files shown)


validator/src/domain/event.rs
```diff
@@ -0,0 +1,35 @@
...
+    UpdateImpressionPrice {
+        publisher: String,
+        price: u64,
+        ad_unit: AdUnit,
+    },
+    ImpressionWithCommission {
+        earner: Vec<Earner>,
+    },
+    ImpressionPricePerCase {},
+    Pay {
```
elpiel: It appears that Pay & PauseChannel & Close?!? also require adUnit in the JS impl?

validator/src/domain/event.rs
```diff
@@ -0,0 +1,35 @@
...
+#[serde(tag = "type", rename_all = "SCREAMING_SNAKE_CASE")]
+#[derive(Serialize, Deserialize)]
+pub enum Event {
+    Impression {
+        publisher: String,
+    },
+    #[serde(rename_all = "camelCase")]
+    UpdateImpressionPrice {
+        publisher: String,
+        price: u64,
```
elpiel: Should it be `BigNum`?

Ivshti: yes

validator/src/domain/event.rs
Ivshti: that's wrong, please PR to the JS impl or discuss with Sam

validator/src/domain/event.rs
```diff
@@ -0,0 +1,40 @@
...
+        ad_unit: AdUnit,
+    },
+    PauseChannel {},
+    #[serde(skip_deserializing)]
+    Close {},
+}
+
+#[derive(Serialize, Deserialize)]
+pub struct Earner {
+    publisher: String,
```
Ivshti: this should be `address`, and we should `#[serde(rename)]` to maintain compat

validator/src/domain/event.rs
elpiel: @samparsky which types should have the `AdUnit`?

validator/src/domain/event.rs
```diff
@@ -0,0 +1,244 @@
...
+use domain::{AdUnit, BalancesMap, BigNum, Channel, ChannelId, DomainError};
+
+#[serde(tag = "type", rename_all = "SCREAMING_SNAKE_CASE")]
+#[derive(Serialize, Deserialize)]
+pub enum Event {
+    #[serde(rename_all = "camelCase")]
+    Impression {
+        publisher: String,
+        // clippy warning for big size difference, because of field
+        ad_unit: Box<AdUnit>,
```
Ivshti: noo that should be an ID

Ivshti: and it should be Option<_>

validator/src/domain/event.rs
```diff
@@ -0,0 +1,244 @@
...
+    accounting: &Accounting,
+    aggregates: &[EventAggregate],
+    channel: &Channel,
+) -> Result<(BalancesMap, Accounting), DomainError> {
+    let deposit = channel.deposit_amount.clone();
+
+    let last_event_aggregate = [accounting.last_event_aggregate]
+        .iter()
+        .chain(aggregates.iter().map(|aggr| &aggr.created))
+        .max_by(|lhs, rhs| lhs.cmp(rhs))
```
Ivshti: why not `max`?

elpiel: You are right, overlooked this. I think it's part of the refactor later on, because I switched it to mapping the datetimes instead of using the aggr.

validator/src/domain/event.rs
elpiel: > noo that should be an ID

Is this the ipfs hash ( the ID ) or something else? This is purely for me, as I still don't know everything and it was even hard to find the structures for the events.

validator/src/domain/event.rs
elpiel: Done

PR #126, opened 2019-08-27
Issue #100 adview manager logic

### LINKED ISSUE #100: adex-adview-manager logic in Rust
it should be in a separate crate, but we need the ad picking logic of adex-adview-manager ported to Rust so that it can be used seamlessly on more platforms (e.g. android, iOS) 

it will reuse the Unit type from the domain crate

we're referring to `applySelection` and `getHTML`

elpiel:
Resolves #100

### DIFF
--- a/adview-manager/src/lib.rs
+++ b/adview-manager/src/lib.rs
@@ -1,0 +1,162 @@
+#![deny(rust_2018_idioms)]
+#![deny(clippy::all)]
+
+use serde::{Deserialize, Serialize};
+use adex_primitives::{AdUnit, TargetingTag, BigNum, SpecValidators};
+use adex_primitives::market_channel::{MarketChannel, MarketStatusType};
+use chrono::Utc;
+
+pub type TargetingScore = f64;
+pub type MinTargetingScore = TargetingScore;
+
+#[derive(Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct AdViewManagerOptions {
+    // Defaulted via defaultOpts
+    #[serde(rename = "marketURL")]
+    pub market_url: String,
+    /// Defaulted
+    pub accepted_states: Vec<MarketStatusType>,
+    /// Defaulted
+    pub min_per_impression: BigNum,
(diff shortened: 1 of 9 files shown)


adview-manager/src/lib.rs
```diff
@@ -0,0 +1,338 @@
...
+            {
+                Some(Unit::new(by_price, targeting_score))
+            } else {
+                None
+            }
+        })
+        .collect();
+    by_score.sort_by(|a, b| {
+        a.targeting_score
+            .partial_cmp(&b.targeting_score)
```
Ivshti: why is this returning a Result? just curious

adview-manager/src/lib.rs
```diff
@@ -0,0 +1,338 @@
...
+            .expect("Should always be comparable")
+    });
+    by_score.truncate(options.top_by_score);
+
+    by_score
+}
+
+fn is_video(ad_unit: &AdUnit) -> bool {
+    ad_unit.media_mime.split('/').take(1).collect::<Vec<&str>>()[0] == "video"
```
Ivshti: you can just `.nth(1) == "video"`

adview-manager/src/lib.rs
elpiel: Targeting score is f64, we can't directly compare them, but we can do a `PartialCmp`:
[Docs](https://doc.rust-lang.org/std/cmp/trait.PartialOrd.html#tymethod.partial_cmp)
```
let result = std::f64::NAN.partial_cmp(&1.0);
assert_eq!(result, None);
```

adview-manager/src/lib.rs
elpiel: Done

elpiel:
Since Clippy was actually kinda disabled and currently the validator worker is the default project + there are quite unfinished stuff that trigger clippy from the "Refactor implementation" I've forgot to run clippy on this new crate. I will do that and fix the warnings!

elpiel:
Fixed @Ivshti

PR #127, opened 2019-08-28
Issue 119 migrate is healthy and is valid transaction

### LINKED ISSUE #119: Refactor: Migrate implmentation of isHealth & isvalidTransition

elpiel:
Resolves #119 
- Only required for those test (the is_healthy & is_valid_transaction tests) fixtures and code was uncommented/re-added
- the HEALTH_THRESHOLD_NEG is passed as function argument as samparsky suggested

### DIFF
--- a/primitives/src/channel_fixtures.rs
+++ b/primitives/src/channel_fixtures.rs
@@ -1,14 +1,16 @@
-//use chrono::{DateTime, Utc};
-//use fake::faker::*;
-//use time::Duration;
-//
-//use crate::asset::fixtures::get_asset;
-//use crate::fixtures::{get_targeting_tags, get_validator};
-//use crate::test_util;
-//use crate::BigNum;
-//
-//use super::{Channel, ChannelId, ChannelSpec, SpecValidators, ValidatorDesc};
-//
+use chrono::{DateTime, Utc};
+use fake::faker::*;
+use time::Duration;
+
+use crate::targeting_tag::fixtures::get_targeting_tags;
+use crate::validator::fixtures::get_validator;
+use crate::BigNum;
+
+use super::{Channel, ChannelSpec, SpecValidators, ValidatorDesc};
+use crate::util::tests::take_one;
+
+const ASSETS_LIST: [&str; 8] = ["DAI", "BGN", "EUR", "USD", "ADX", "BTC", "LIT", "ETH"];
+
 ///// It will get the length of channel_id bytes and will fill enough bytes in front
 ///// If > 32 bytes &str is passed it will `panic!`
 //pub fn get_channel_id(channel_id: &str) -> ChannelId {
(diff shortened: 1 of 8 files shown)


validator_worker/src/core/follower_rules.rs
```diff
@@ -4,51 +4,58 @@ pub fn is_valid_transition(channel: &Channel, prev: &BalancesMap, next: &Balance
...
         .sum();
 
-     if sum_approved_mins >= sum_our {
+    if sum_approved_mins >= sum_our {
         return true;
     }
 
-     let deposit = &channel.deposit_amount;
+    let deposit = &channel.deposit_amount;
     let acceptable_difference = deposit * &health_threshold_neg / &BigNum::from(1_000);
```
samparsky: I think this should be 👇 where `health_threshold` is the user configuration health threshold value that is passed into the pure function
``` 
let health_threshold_neg: BigNum = (1_000 - health_threshold).into();	
```

primitives/src/channel_fixtures.rs
```diff
@@ -67,53 +68,52 @@
...
+    Pair {
+        leader: ValidatorDesc,
+        follower: ValidatorDesc,
+    },
+    SpecValidators(SpecValidators),
+    Generate {
+        validators_prefix: &'a str,
+    },
+}
+
```
samparsky: There seems to be no need for this. If we want to generate validators for fixtures it would be more constrained around `fees` and not `id` or `url`. In fact, the `URL` & `id` for fixtures should be hardcoded similar to the JS implementation else we would have issues when running tests as the validator id wouldn't be found in the spec.

primitives/src/validator.rs
```diff
@@ -104,3 +106,22 @@ pub enum MessageTypes {
...
+    use fake::faker::*;
+
+    use crate::BigNum;
+
+    use super::ValidatorDesc;
+
+    pub fn get_validator<V: AsRef<str>>(validator_id: V, fee: Option<BigNum>) -> ValidatorDesc {
+        let fee = fee.unwrap_or_else(|| BigNum::from(<Faker as Number>::between(1, 13)));
+        let url = format!(
+            "http://{}-validator-url.com/validator",
```
samparsky: The URL would lead to a lot of errors when the validator tick is running because it will try to post heartbeat messages to the sentry node specified in this URL.
We should stick to the URL in the JS stack implementation i.e. 
```
validators: [
				{ id: 'awesomeLeader', url: 'http://localhost:8005', fee: '100' },
				{ id: 'awesomeFollower', url: 'http://localhost:8006', fee: '100' },
			]
```

primitives/src/channel_fixtures.rs
```diff
@@ -67,53 +68,52 @@
...
+    let validators = match validators_option {
+        ValidatorsOption::Pair { leader, follower } => [leader, follower].into(),
+        ValidatorsOption::SpecValidators(spec_validators) => spec_validators,
+        ValidatorsOption::Generate { validators_prefix } => [
+            get_validator(&format!("{} leader", validators_prefix), None),
+            get_validator(&format!("{} follower", validators_prefix), None),
+        ]
+        .into(),
+    };
```
samparsky: Here we can just have

```
 let validators = [
 get_validator(&format!("{} leader", validators_prefix), None),
 get_validator(&format!("{} leader", validators_prefix), None),
]
```
No need for the url to change so maybe a `fee` param only

primitives/src/channel_fixtures.rs
```diff
@@ -1,14 +1,16 @@
...
+use time::Duration;
+
+use crate::targeting_tag::fixtures::get_targeting_tags;
+use crate::validator::fixtures::get_validator;
+use crate::BigNum;
+
+use super::{Channel, ChannelSpec, SpecValidators, ValidatorDesc};
+use crate::util::tests::take_one;
+
+const ASSETS_LIST: [&str; 8] = ["DAI", "BGN", "EUR", "USD", "ADX", "BTC", "LIT", "ETH"];
```
samparsky: FYI the`depositAsset` would be an Eth address in production

elpiel: We do need some dummy data, so I am opened for suggestions to improve this.

samparsky: Yeah it's fine I was just pointing it out, because its not apparent at first glance

primitives/src/channel_fixtures.rs
elpiel: `fee` param to which one?

primitives/src/channel_fixtures.rs
elpiel: It's not apparent with the `String` and no documentation comments on the `Channel` as well

PR #132, opened 2019-09-02
Sentry implementation

### LINKED ISSUE #124: Refactor: Implement all methods sentry interface

samparsky:
Fixes #124


primitives/src/channel.rs
```diff
@@ -38,10 +38,14 @@ pub struct ChannelSpec {
...
+    pub event_submission: Option<EventSubmission>,
     /// A millisecond timestamp of when the campaign was created
-    #[serde(with = "ts_milliseconds")]
-    pub created: DateTime<Utc>,
+    #[serde(
+        default,
+        skip_serializing_if = "Option::is_none",
+        with = "ts_milliseconds_option"
+    )]
+    pub created: Option<DateTime<Utc>>,
```
elpiel: Isn't this required?

primitives/src/channel.rs
```diff
@@ -51,7 +55,7 @@ pub struct ChannelSpec {
     )]
     pub active_from: Option<DateTime<Utc>>,
     /// A random number to ensure the campaignSpec hash is unique
-    pub nonce: BigNum,
+    pub nonce: Option<BigNum>,
```
elpiel: Isn't this also required? I see that in the js impl there is no `required()`, but just to check as we don't have this documented.

primitives/src/config.rs
```diff
@@ -3,60 +3,8 @@ use serde::{Deserialize, Serialize};
...
-    ethereum_core_address = '0x333420fc6a897356e69b62417cd17ff012177d2b'
-    ethereum_network = 'homestead'
-    token_address_whitelist = ['0x89d24A6b4CcB1B6fAA2625fE562bDD9a23260359']
-
-    creators_whitelist = []
-    minimal_deposit = 0
-    minimal_fee = 0
-    validators_whitelist = []
-    "#;
+pub const DEVELOPMENT_CONFIG: &str = include_str!("../../docs/config/dev.toml");
```
elpiel: can't you parse the tomls directly and use the structs instead of `&str`s?

primitives/src/config.rs
```diff
@@ -71,7 +19,7 @@ pub struct RateLimit {
 
 #[derive(Serialize, Deserialize, Debug, Clone)]
 pub struct Config {
-    pub identity: String, // should not be here maybe?
+    pub identity: Option<String>, // should not be here maybe?
```
elpiel: What's the point of not having an identity?

validator_worker/src/sentry_interface.rs
```diff
@@ -1,143 +1,265 @@
...
-        .and_then(|mut res: Response| res.json::<ChannelAllResponse>());
-
-    await!(future.compat())
-}
-
-
-#[derive(Clone)]
-pub(crate) struct SentryApi {
+#[derive(Clone, Debug)]
+pub struct SentryApi<T: Adapter> {
```
elpiel: Is this acting like a service or is it storing data? I am confused.
If you want it to be a service, then it should be stateless, i.e. no `validator` and `channel`. Since the `adapter`, `sentry_url` and etc. can be used for any `validator`+`channel` and then don't reveal state.

validator_worker/src/sentry_interface.rs
```diff
@@ -1,143 +1,265 @@
...
+            .timeout(Duration::from_secs(config.fetch_timeout.into()))
+            .build()
+            .unwrap();
+
+        match validator {
+            Some(v) => {
+                let sentry_url = format!("{}/channel/{}", v.url, channel.id);
+
+                Ok(Self {
+                    validator: v.clone().to_owned(),
```
elpiel: why `clone` + `to_owned`?

validator_worker/src/sentry_interface.rs
```diff
@@ -1,143 +1,265 @@
...
+            }
+            None => Err(ValidatorWorkerError::InvalidValidatorEntry(
+                "we can not find validator entry for whoami".to_string(),
+            )),
         }
     }
 
-    pub fn propagate() {
+    pub fn propagate(&self, messages: Vec<MessageTypes>) {
+        let serialised_messages: Vec<String> = messages
```
elpiel: why serialize the message here and pass a String?

validator_worker/src/sentry_interface.rs
```diff
@@ -1,143 +1,265 @@
...
+            .and_then(|mut res: Response| res.json::<LastApprovedResponse>());
 
         await!(future.compat())
     }
 
-    pub fn get_last_msgs() {
-         let future = self
+    pub async fn get_event_aggregates(
+        &self,
+        after: Option<u32>,
```
elpiel: what's this after u32? Can you be more specific? I don't thin this is really clear for somebody reading the code.

primitives/src/channel.rs
samparsky: Not quite

primitives/src/channel.rs
samparsky: Not at all

primitives/src/channel.rs
elpiel: Don't we always set it when channel is created?

Ivshti: we do, it should not be an option

primitives/src/config.rs
Ivshti: again, shouldn't be an option

validator_worker/src/sentry_interface.rs
Ivshti: it's basically "get all event aggregates AFTER this point" 

but it looks like it should be a DateTime, and not optional (cause you can always pass a zero timedate, 1970)

primitives/src/config.rs
Ivshti: it will need lazy_static I think

primitives/src/config.rs
samparsky: I think identity shouldn't be part of `Config` becuase it's not a configuration hence the option<String> to make the code compile

primitives/src/config.rs
elpiel: For sure, already did that with `prep_db`

validator_worker/src/sentry_interface.rs
samparsky: cos it's much easier and you don't have to serialized messages in both the leader and follower

validator_worker/src/sentry_interface.rs
samparsky: No that is incorrect, the sentry_url could be different from the validator_url. I have made a change to the name to make it distinct. A validator_url is specified in the channel spec while the sentry_url is passed a flag during startup

samparsky: Also a Sentry is channel specific hence adding the `channel` property

validator_worker/src/sentry_interface.rs
samparsky: Removed

validator_worker/src/sentry_interface.rs
samparsky: yeah made changed it to Datetime as its more descriptive

primitives/src/config.rs
samparsky: I made the change to `lazy_static`

Ivshti: now that I think about it, lazy_static is perhaps not the best idea, cause what happens if parsing fails? We blow up without an ability to handle as we want

validator_worker/src/sentry_interface.rs
Ivshti: @elpiel it's not storing data, but it's constructed with sentry_url and channel as context

validator_worker/src/sentry_interface.rs
```diff
@@ -1,143 +1,264 @@
...
-pub(crate) struct SentryApi {
-    pub sentry_url: String,
+#[derive(Clone, Debug)]
+pub struct SentryApi<T: Adapter> {
+    pub adapter: T,
+    pub validator_url: String,
     pub client: Client,
     pub logging: bool,
-    pub channel: Channel
+    pub channel: Channel,
```
elpiel: I'm still not sure why won't you pass the channel to the functions that need it, instead of having it on instantiation of the struct.
E.g. in the methods: get_latest_msg, get_last_msgs, propagate and etc. you don't need it. They need identity - they have it. The functions requiring the channel to have validator with `whoami` value should handle this on their own IMO. Otherwise we are blocking these generic non-channel dependant functions to require a channel. At the end we might end-up using an instance of 1 SentryApi with 1 channel to do work for totally different channel(for example batch processing).

validator_worker/src/sentry_interface.rs
```diff
@@ -1,143 +1,264 @@
...
+                Err(e) => handle_http_error(e, &validator.url),
+            }
+        }
     }
 
-    async pub fn get_latest_msg(from: &str, type: &str) -> impl Future<Output = Result<Vec<Channel>, Error>>  {
+    pub async fn get_latest_msg(
+        &self,
+        from: String,
+        message_type: String,
```
elpiel: This won't work with the Enum, that's why I had the complicated code for the `Message` enum in pre-refactor code.

validator_worker/src/sentry_interface.rs
```diff
@@ -1,143 +1,264 @@
...
+        all.push(first_page);
+        let result_all: Vec<Channel> = all
+            .into_iter()
+            .flat_map(|ch| ch.channels.into_iter())
+            .collect();
+        Ok(result_all)
+    }
+}
+
+pub async fn fetch_page(
```
elpiel: We need to be careful what we expose. fetch_page is not part of the public API as it's only used in the `SentryApi`.
I saw quite some places like that so we need to review our API soon.

validator_worker/src/sentry_interface.rs
elpiel: But you are passing the serialized message to `propagate_to` which is not public? This has nothing to do with the `leader`/`follower`, I just meant that you can use the MessageType and encode it when you are sending the request, not before that and loosing all the type-safety stuff

validator_worker/src/sentry_interface.rs
Ivshti: You have a point.

We should rethink the Sentry API interface; perhaps channel should not be used in instantiation, but rather for the methods that need it

validator_worker/src/sentry_interface.rs
Ivshti: perhaps that's a problem of the documentation? pages should be documented IMO

elpiel: It is meant as inner fn for `all_channels()` only, which should be public API itself(as it is) and I guess we can later even move it to the `SentryApi`.
But for `fetch_page` not sure if should be exposed if we need it only for `all_channels`?

primitives/src/config.rs
samparsky: Yeah I want it to blow up, because that's the default configuration from Adex so if it's not correct then it should blow up

validator_worker/src/sentry_interface.rs
samparsky: The channel is needed in instantiation to get the channel validator url and validate that the validator is meant to be validating the channel.
Also depending on how the sentry interface grows there could definitely be a need for other channel properties in the future.
Also, Sentry API is channel dependent ( i.e. a Sentry API is for one channel, so for multiple channels, there would be multiple Sentry API's )

primitives/src/config.rs
Ivshti: that makes sense, cool; just make sure to use .expect rather than .unwrap

primitives/src/channel.rs
samparsky: No, we don't https://github.com/AdExNetwork/adex-validator/blob/master/routes/channelCreate.js#L20

When a channel is created is not a much needed detail. What are majorly required is `validUntil` and `withdrawPeriodStart`

validator_worker/src/sentry_interface.rs
samparsky: hmnnn ... do we really need an enum? it could be just string constants

validator_worker/src/sentry_interface.rs
samparsky: There is no losing type safety stuff. The propagate function is public and it has the appropriate type definition. The propagate_to function does one thing only takes in an array of encoded messages and posts it a validators url

validator_worker/src/sentry_interface.rs
elpiel: I guess there are multiple way to work on this issue indeed.
We might not check if the identity is in the channel inside `new`, but instead where we work with the channel. The rest of the methods don't deal with the Channel either way and if you try to get messages for an identity that doesn't exist - that's ok, right?

We might separate this? Doing fetching of aggregates, channels, messages + posting some messages and etc. isn't it too much for 1 object?
I am trying to also figure that out myself.

samparsky: @elpiel There is no need for separating this. They all point to the same resource.
Sentry API is channel dependent, each of those methods actually uses the channel in an abstracted way i.e. validator_url

validator_worker/src/sentry_interface.rs
elpiel: You could, but then you need to match the string and the signature is not verbose enough ( you need to specify in the documentation that this string should be one of the values in those constants or those string literals).
Plus if you forget to add the match statement for a new `MessageType`, and just a note that this can be 1 of many places that we use a `String` for the type, in all places you'll need to match the string and handle all cases

elpiel: I was trying to figure out the same thing:
https://github.com/AdExNetwork/adex-validator-stack-rust/blob/master/domain/src/validator/message.rs#L45

And it got ugly, but I couldn't find anything in serde that will help with this, maybe I've missed something.

Ivshti: there's no solution other than to build two separate enums... or use a String

elpiel: I am throwing it just as an option I found:
https://crates.io/crates/enum-kinds
But indeed it's ugly and I might check in serde(i.e. ask) for feature like that or if they will help me to do it in my free time, until then String I guess..

validator_worker/src/sentry_interface.rs
elpiel: So let me get this straight, you want the `Heartbeat`/`producer`/`leader`/`follower` fnunctions to get a `SentryApi` instance which is bound to a channel and instantiate it before calling those functions?

samparsky: I think you should review the JS implementation. The sentry API is channel dependent and so is the heartbeat, producer, leader, follower

elpiel: Yeah, but I mean if you have the Channel in the SentryApi, then you don't need to pass it to any of the heartbeat, producer, leader, follower, you only need SentryApi

validator_worker/src/sentry_interface.rs
```diff
@@ -1,143 +1,261 @@
...
-        self.get_latest_msg(whoami, type)
+    pub async fn get_our_latest_msg(
+        &self,
+        message_type: String,
+    ) -> Result<ValidatorMessageResponse, reqwest::Error> {
+        let whoami = self.adapter.whoami();
+        await!(self.get_latest_msg(whoami, message_type))
+    }
+
+    pub async fn get_last_approved(&self) -> Result<LastApprovedResponse, reqwest::Error> {
```
elpiel: Since I am working branched out of this branch, I found out that this is incorrect as `LastApprovedResponse` has a `Vec` of messages, where as this method returns single message. I will probably fix it, but I am noting it for reference.

validator_worker/src/sentry_interface.rs
```diff
@@ -1,143 +1,261 @@
...
+            .and_then(|mut res: Response| res.json::<LastApprovedResponse>());
 
         await!(future.compat())
     }
 
-    pub fn get_last_msgs() {
-         let future = self
+    pub async fn get_event_aggregates(
+        &self,
+        after: DateTime<Utc>,
```
elpiel: Now I am not sure how correct is the JS implementation and how it makes sense here, but in the Producer if there is no previous `Accounting` message, it will create one with `lastEvAggr: new Date(0)` (i.e. `Jan 01 1970`) and it will use it for the `after`. Maybe we do need the after to be an Option after all?

@Ivshti @samparsky

validator_worker/src/sentry_interface.rs
samparsky: I don't think that's correct

samparsky: I think ApproveState returns Vec<ApproveState> and NewState returns a single message

elpiel: Ok, so I am apparently missing something?
https://github.com/AdExNetwork/adex-validator/blob/master/routes/channel.js#L167

validator_worker/src/sentry_interface.rs
elpiel: this is a question.

validator_worker/src/sentry_interface.rs
elpiel: this is also an opened question:
https://github.com/AdExNetwork/adex-validator/blob/master/routes/channel.js#L190

Issue #84, opened 2019-06-24
Validator MerkleTree


elpiel:
I've created a test MerkleTree, to see how to implement it, since we need it for the Heartbeat for example.

- Test - The MerkleTree hasn't been tested though, so check if this implementation will work.
- add to separate crate
- add the crate to `validator`

http://github.com/elpiel/merkle-tree

PR #134, opened 2019-09-02
Merkle implementation

samparsky:
Fix #96

### DIFF
--- a/primitives/Cargo.toml
+++ b/primitives/Cargo.toml
@@ -29,5 +29,6 @@ fake = { version = "^1.3", features = ["chrono"] }
 rand = { version = "^0.6" }
 lazy_static = "1.4.0"
 merkletree = "0.10.0"
+merkle = "1.11.0"
 rust-crypto = "^0.2"
 tiny-keccak = "1.5"
--- a/primitives/Cargo.toml
+++ b/primitives/Cargo.toml
@@ -28,3 +28,6 @@ num-derive = "0.2"
 fake = { version = "^1.3", features = ["chrono"] }
 rand = { version = "^0.6" }
 lazy_static = "1.4.0"
+merkletree = "0.10.0"
+rust-crypto = "^0.2"
+tiny-keccak = "1.5"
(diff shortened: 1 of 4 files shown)


primitives/Cargo.toml
```diff
@@ -27,4 +27,7 @@ num-derive = "0.2"
 # Fixtures
 fake = { version = "^1.3", features = ["chrono"] }
 rand = { version = "^0.6" }
-lazy_static = "1.4.0"
\ No newline at end of file
+lazy_static = "1.4.0"
+merkletree = "0.10.0"
+rust-crypto = "^0.2"
```
elpiel: Why did you choose `rust-crypto`? There is stand alone `sha3` crate + it also includes `Keccak`. But if you have a good reason about using `tiny_keccak` I will be ok with that.

But for Sha3 I don't think we need the whole `rust-crypto`

primitives/src/merkle_tree.rs
```diff
@@ -0,0 +1,99 @@
...
+    fn reset(&mut self) {
+        self.0 = Keccak::new_sha3_256();
+    }
+}
+
+pub struct MerkleTree(merkle::MerkleTree<MerkleItem, KeccakAlgorithm, VecStore<MerkleItem>>);
+
+impl MerkleTree {
+    pub fn new(leaves: Vec<MerkleItem>) -> MerkleTree {
+        let t = merkle::MerkleTree::from_iter(leaves);
```
elpiel: If you followed on my comment in the issue:
https://github.com/AdExNetwork/adex-validator-stack-rust/issues/84#issuecomment-510789831
The main issue I've bumped in to was that is asserts that there are more than 1 leaves, which means that the heartbeat MerkleTree we need to create won't be possible:

https://docs.rs/merkletree/0.10.0/src/merkletree/merkle.rs.html#915

https://github.com/AdExNetwork/adex-validator/blob/a3b59a4c43d02ab4fc9de8ed5c48957592a260d4/services/validatorWorker/heartbeat.js#L13

samparsky: Actually the entire implementation is wrong, including the previous one. The merkle tree implementation doesn't support 8byte arrays of 64 length which is the size of nodes/leafs in the JS implementation.

PR #137, opened 2019-09-10
Issue #122 Heartbeat & testing get_signable_state_root

### LINKED ISSUE #122: Refactor: Heartbeat implementation

elpiel:
Resolves #122


elpiel:
ping @samparsky

validator_worker/src/heartbeat.rs
```diff
@@ -1,146 +1,66 @@
...
+    iface: &SentryApi<A>,
+    balances: BalancesMap,
+    heartbeat_time: u32,
+) -> Result<(), Box<dyn Error>> {
+    let validator_message_response = await!(iface.get_our_latest_msg("Heartbeat".into()))?;
+
+    let heartbeat_msg = validator_message_response
+        .msg
+        .get(0)
+        .and_then(|message_types| match message_types {
```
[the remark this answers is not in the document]
elpiel: if map is used we need to also use `.transponse` as we get an `Option<Option<Heartbeat>>`

adapter/src/lib.rs
```diff
@@ -4,6 +4,7 @@
 #![deny(clippy::match_bool)]
 #![doc(test(attr(feature(async_await, await_macro))))]
 #![doc(test(attr(cfg(feature = "dummy-adapter"))))]
+#![allow(dead_code)]
```
elpiel: whoops that slipped away, will remove.

adapter/src/lib.rs
```diff
@@ -164,13 +164,13 @@ fn encode_params(
 
     let params: Vec<_> = types
```
elpiel: Probably this can be quite simplified and probably we can use `&str`, but it's not a job for this PR.

adapter/src/lib.rs
```diff
@@ -186,3 +186,35 @@ fn parse_tokens(params: &[(ParamType, &str)], lenient: bool) -> Result<Vec<Token
...
+#[cfg(test)]
+mod test {
+    use super::*;
+    use byteorder::{BigEndian, ByteOrder};
+    use chrono::{TimeZone, Utc};
+    use primitives::merkle_tree::MerkleTree;
+    use std::convert::TryFrom;
+
+    #[test]
+    fn test_get_signable_state_root_hash_is_aligned_with_js_impl() {
```
elpiel: I've decided to test with the Heartbeat state root, that's also why I've included the byteorder ( I will move it to dev-dependencies)

PR #139, opened 2019-09-12
Merkle implementation

samparsky:
Fixes #134


primitives/src/merkle_tree.rs
```diff
@@ -0,0 +1,186 @@
...
+    fn node(&mut self, left: MerkleItem, right: MerkleItem, _height: usize) -> MerkleItem {
+        let left_vec: Vec<u8> = left.to_vec();
+        let right_vec: Vec<u8> = right.to_vec();
+
+        let mut node_vec = vec![left_vec, right_vec];
+        node_vec.sort();
+
+        let flatten_node_vec: Vec<u8> = node_vec.into_iter().flatten().collect();
+
+        self.write(&flatten_node_vec.as_slice());
```
elpiel: I think you can just do `self.write(&flatten_node_vec);` the Vec implements the same thing as the String, so it should Deref it to a slice

primitives/src/merkle_tree.rs
```diff
@@ -0,0 +1,186 @@
...
+    tree: Tree,
+    root: MerkleItem,
+}
+
+impl MerkleTree {
+    pub fn new(data: &[MerkleItem]) -> MerkleTree {
+        let mut leaves: Vec<MerkleItem> = data.to_owned();;
+
+        let tree: Tree = if leaves.len() == 1 {
+            Tree::SingleItem(leaves.first().unwrap().to_owned())
```
elpiel: just a suggestion: `leaves[0].to_owned()` can be a bit cleaner.

primitives/src/merkle_tree.rs
```diff
@@ -0,0 +1,186 @@
...
+        } else {
+            // sort the merkle tree leaves
+            leaves.sort();
+            // remove duplicates
+            leaves.dedup_by(|a, b| a == b);
+
+            let merkletree = merkle::MerkleTree::from_iter(leaves.clone());
```
elpiel: is this `.clone()` necessary? You should be able to just move the value inside. The current fn owns it and it's not used afterwards.

primitives/src/merkle_tree.rs
```diff
@@ -0,0 +1,186 @@
...
+    }
+}
+
+#[cfg(test)]
+mod test {
+    use super::*;
+    use hex::FromHex;
+
+    #[test]
+    fn it_works_okay_with_js_impl() {
```
elpiel: Please rename this to something more relevant.

primitives/src/merkle_tree.rs
```diff
@@ -0,0 +1,186 @@
...
+        let top = MerkleTree::new(&[h1, h2]);
+
+        let root = hex::encode(&top.root());
+
+        assert_eq!(
+            root, "70d6549669561c65fdc687b87743b67e494e1f4be5d19a2955507220e57baaa6",
+            "should generate the correct root"
+        );
+
+        let proof = top.proof(0);
```
elpiel: `fn proof()` returns owned data, not a reference, so the `clone()` bellow is unnecessary:
`let verify = top.verify(proof);`

primitives/src/merkle_tree.rs
```diff
@@ -0,0 +1,186 @@
...
+        let verify = top.verify(proof.clone());
```
elpiel: same goes for this `.clone()`

primitives/src/merkle_tree.rs
```diff
@@ -0,0 +1,186 @@
...
+        );
+
+        let proof = top.proof(0);
+
+        let verify = top.verify(proof.clone());
+        assert_eq!(verify, true, "should verify proof successfully");
+    }
+
+    #[test]
+    fn it_works_okay_with_duplicate_leaves_js_impl() {
```
elpiel: Please, do not postfix with `js_impl`.

PR #140, opened 2019-09-16
Issues #118 Leader/follower

### LINKED ISSUE #118: Refactor: Implement Leader & Follower Tick

elpiel:
Resolves #118


primitives/src/util/tests/prep_db.rs
```diff
@@ -11,9 +11,10 @@ lazy_static! {
         ids.insert("leader", "awesomeLeader".into());
         ids.insert("follower", "awesomeFollower".into());
         ids.insert("user", "awesomeTestUser".into());
-        ids.insert("publisher", "myAwesomePublisher".into());
+        ids.insert("publisher", "b7d3f81e857692d13e9d63b232a90f4a1793189e".into());
```
elpiel: I've added real Hashes taken from `Jerry`, since Rust is typesafe and explodes when you pass anything that is invalid to the ABI crate for Address, FixedBytes and etc.

primitives/src/validator.rs
```diff
@@ -62,7 +62,7 @@ pub struct RejectState {
     pub reason: String,
     pub state_root: String,
     pub signature: String,
-    pub balances: Option<String>,
+    pub balances: Option<BalancesMap>,
```
elpiel: We should think about what is really necessary for RejectState and how to actually add it, as these Options don't tell me anything at this point:
- When should they be present?
- Why is the timestamp an Option?
- In what cases we do have RejectState?

validator_worker/src/follower.rs
```diff
@@ -1,12 +1,139 @@
...
+use crate::{get_state_root_hash, producer};
+use chrono::Utc;
+
+enum InvalidNewState {
+    RootHash,
+    Signature,
+    Transition,
+}
+
+enum NewStateResult {
```
elpiel: this is internal `Result` type, as @Ivshti mentioned, things are error because invalid transactions, shouldn't implement Error.
Atm if we use Result, instead of custom Error enum, we will need to handle the Error somehow and I am not sure if this is sufficient and if I understood @Ivshti correct.

validator_worker/src/follower.rs
```diff
@@ -1,12 +1,139 @@
...
+    )? {
+        return Ok(await!(on_error(
+            &iface,
+            &new_state,
+            InvalidNewState::Signature
+        )));
+    }
+
+    let last_approve_response = await!(iface.get_last_approved())?;
+    // TODO: Check if it's possible to have an empty response (i.e. we don't have LastApprove)
```
elpiel: See https://github.com/AdExNetwork/adex-validator/blob/master/services/validatorWorker/follower.js#L45

validator_worker/src/follower.rs
```diff
@@ -1,12 +1,139 @@
...
+    .to_string();
+
+    iface.propagate(&[&MessageTypes::RejectState(RejectState {
+        reason,
+        state_root: new_state.state_root.clone(),
+        signature: new_state.signature.clone(),
+        balances: Some(new_state.balances.clone()),
+        /// The NewState timestamp that is being rejected
+        // TODO: Double check this, if we decide to have 2 timestamps - 1 for the RejectState & 1 for NewState timestamp
+        timestamp: Some(Utc::now()),
```
elpiel: Where should this come from?!

samparsky: This should come from the rejected message

validator_worker/src/follower.rs
samparsky: last_approved can have an empty response.
It's a JS check against "{}" so in the case of Rust it would be None

primitives/src/util/tests/prep_db.rs
samparsky: 👍

primitives/src/validator.rs
samparsky: RejectState cases -> InvalidSignature, InvalidRootHash etc
The properties of RejectState depend on the State being rejected.

elpiel: If it depends on the State being rejected, why don't we just include the whole message, instead of making variants with `Option` and then in all the code check for those variants?

validator_worker/src/leader.rs
```diff
@@ -1,12 +1,37 @@
...
+}
+
+fn on_new_accounting<A: Adapter + 'static>(
+    iface: &SentryApi<A>,
+    (balances, new_accounting): (&BalancesMap, &Accounting),
+) -> Result<(), Box<dyn Error>> {
+    let state_root_raw = get_state_root_hash(&iface, &balances)?;
+    let state_root = hex::encode(state_root_raw);
+
+    let signature = iface.adapter.sign(&hex::encode(&state_root))?;
```
samparsky: you have double hex::encode() here, should be one

validator_worker/src/lib.rs
```diff
@@ -3,19 +3,109 @@
...
+}
+
+#[cfg(test)]
+mod test {
+    use super::*;
+
+    use adapter::DummyAdapter;
+    use primitives::adapter::AdapterOptions;
+    use primitives::util::tests::prep_db::{DUMMY_CHANNEL, IDS};
+    use primitives::{config::DEVELOPMENT_CONFIG, BalancesMap, Channel};
```
samparsky: There is no need DEVELOPMENT_CONFIG this we can use configuration function in config
https://github.com/AdExNetwork/adex-validator-stack-rust/blob/dev/primitives/src/config.rs#L57

validator_worker/src/follower.rs
elpiel: My point: It is ambiguous. It _can_ be the state being rejected timestamp, but usually it means the timestamp of the RejectState.

validator_worker/src/follower.rs
elpiel: So we have:
```
pub struct LastApprovedResponse {
    pub last_approved: Option<LastApproved>, <---- This?
    pub heartbeats: Option<Vec<Heartbeat>>,
}
```

And do we have this?

```
pub struct LastApproved {
    pub new_state: Option<NewState>, <--- Option here
    /// ApproveState can be None if the channel is brand new and we have only NewState atm
    pub approved_state: Option<ApproveState>, <--- And as we discussed, this is an edge case so we do have this
}
```

validator_worker/src/lib.rs
```diff
@@ -3,19 +3,109 @@
...
+
+    fn setup_iface(channel: &Channel) -> SentryApi<DummyAdapter> {
+        let adapter_options = AdapterOptions {
+            dummy_identity: Some(IDS["leader"].clone()),
+            dummy_auth: None,
```
samparsky: You should provide dummy_auth and dummy_auth_tokens else the tests will fail when the `fix dummy adapter and ethereum adapter` PR is merged

primitives/src/validator.rs
samparsky: i think we should follow the implementation for now.
Later we can create Adex improvement proposals for that

validator_worker/src/follower.rs
samparsky: Yes this is correct.

validator_worker/src/follower.rs
samparsky: i think we should follow the implementation for now.
Later we can create Adex improvement proposals for this

primitives/src/config.rs
```diff
@@ -30,7 +30,7 @@ pub struct Config {
     pub max_channels: u32,
     pub wait_time: u32,
     pub aggr_throttle: u32,
-    pub heartbeat_time: u32,
+    pub heartbeat_time: u32, // in milliseconds
```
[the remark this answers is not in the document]
elpiel: @Ivshti asked me and I couldn't tell from seeing `u32` so I've just added this comment to be more obvious than `u32`

validator_worker/src/leader.rs
elpiel: My bad, thanks!

adapter/src/lib.rs
```diff
@@ -149,50 +147,31 @@ impl EthereumChannel {
...
-
-    let tokens = parse_tokens(&params, lenient)?;
-    let result = encode(&tokens);
-
-    Ok(result.to_vec())
-}
-
-fn parse_tokens(params: &[(ParamType, &str)], lenient: bool) -> Result<Vec<Token>, Box<dyn Error>> {
-    params
+fn encode_params(params: &[(ParamType, &str)], lenient: bool) -> Result<Vec<u8>, Box<dyn Error>> {
```
samparsky: nice improvement! 👍

validator_worker/src/lib.rs
elpiel: Fixed.

validator_worker/src/leader.rs
elpiel: Fixed.

primitives/src/config.rs
```diff
@@ -5,7 +5,7 @@ use std::fs;
 use toml;
 
 lazy_static! {
-    static ref DEVELOPMENT_CONFIG: Config =
+    pub static ref DEVELOPMENT_CONFIG: Config =
```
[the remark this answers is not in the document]
elpiel: Fixed.

PR #143, opened 2019-09-19
Finish Ethereum and Dummy adapter

samparsky:
#142


adapter/contract/AdExCore.json
```diff
@@ -0,0 +1 @@
...e","type":"uint256"}],"name":"channelWithdraw","outputs":[],"payable":false,"stateMutability":"nonpayable","type":"function"},{"constant":false,"inputs":[{"components":[{"name":"creator","type":"address"},{"name":"tokenAddr","type":"address"},{"name":"tokenAmount","type":"uint256"},{"name":"validUntil","type":"uint256"},{"name":"validators","type":"address[]"},{"name":"spec","type":"bytes32"}],"name":"channel","type":"tuple"}],"name":"channelWithdrawExpired","outputs":[],"payable":false,"stateMutability":"nonpayable","type":"function"},{"constant":true,"inputs":[{"name":"","type":"bytes32"}],"name":"states","outputs":[{"name":"","type":"uint8"}],"payable":false,"stateMutability":"view","type":"function"},{"anonymous":false,"inputs":[{"indexed":true,"name":"channelId","type":"bytes32"}],"name":"LogChannelOpen","type":"event"},{"anonymous":false,"inputs":[{"indexed":true,"name":"channelId","type":"bytes32"},{"indexed":false,"name":"amount","type":"uint256"}],"name":"LogChannelWithdrawExpired","type":"event"},{"anonymous":false,"inputs":[{"indexed":true,"name":"channelId","type":"bytes32"},{"indexed":false,"name":"amount","type":"uint256"}],"name":"LogChannelWithdraw","type":"event"}]
```
Ivshti: can't you include adex-protocol-eth as a submodule?

adapter/src/dummy.rs
```diff
@@ -1,7 +1,7 @@
 #![deny(clippy::all)]
 #![deny(rust_2018_idioms)]
```
Ivshti: why is this needed?

adapter/src/dummy.rs
```diff
@@ -63,27 +63,43 @@ impl Adapter for DummyAdapter {
...
+            Ok(_) => Ok(true),
+            Err(e) => Err(AdapterError::InvalidChannel(format!("{}", e))),
+        }
     }
 
-    fn session_from_token(&self, _token: &str) -> AdapterResult<String> {
-        // @TODO
-        Ok("hello".to_string())
+    fn session_from_token(&mut self, token: &str) -> AdapterResult<Session> {
+        let mut identity = "";
```
Ivshti: you should use a functional style here
`let identity = self.tokens_for_auth.iter().find(|(key, val)| val == token)`

adapter/src/dummy.rs
```diff
@@ -63,27 +63,43 @@ impl Adapter for DummyAdapter {
...
+    fn get_auth(&mut self, _validator: &ValidatorDesc) -> AdapterResult<String> {
+        let who = self
+            .tokens_verified
+            .clone()
+            .into_iter()
+            .find(|(_, id)| *id == self.identity);
+
+        match who {
+            Some((id, _)) => {
+                let auth = self.tokens_for_auth.get(&id).unwrap();
```
Ivshti: this `unwrap` should be `expect` so that you can explain why you make that assumption

adapter/src/ethereum.rs
```diff
@@ -1,101 +1,486 @@
...
+                ))
+            }
+        };
+
+        let account = SafeAccount::from_file(
+            serde_json::from_reader(json_file).expect("Failed to read json file"),
+            None,
+            &Some(self.keystore_pwd.clone()),
+        )
+        .expect("Failed to create account");
```
Ivshti: why doesn't this just return an `Err`?

adapter/src/ethereum.rs
```diff
@@ -1,101 +1,486 @@
...
+            Ok(data) => data,
+            Err(_) => {
+                return Err(AdapterError::Configuration(
+                    "Invalid keystore location provided".to_string(),
+                ))
+            }
+        };
+
+        let account = SafeAccount::from_file(
+            serde_json::from_reader(json_file).expect("Failed to read json file"),
```
Ivshti: this should also propagate an `Err` IMHO

adapter/src/ethereum.rs
```diff
@@ -1,101 +1,486 @@
...
-        self.address.unwrap().to_string()
+    fn whoami(&self) -> AdapterResult<String> {
+        match &self.wallet {
+            Some(wallet) => {
+                let address = format!(
+                    "{:?}",
+                    public_to_address(
+                        &wallet
+                            .public(&self.keystore_pwd)
+                            .expect("failed to get public key")
```
Ivshti: same...err handling should use `Err`

adapter/src/ethereum.rs
```diff
@@ -1,101 +1,486 @@
...
+
+                let contract_address =
+                    Address::from_slice(self.config.ethereum_core_address.as_bytes());
+
+                let contract = Contract::from_json(
+                    web3.eth(),
+                    contract_address,
+                    include_bytes!("../contract/Identity.json"),
+                )
+                .unwrap();
```
Ivshti: this is OK to be unwrapped, but you should use `lazy_static` for this

adapter/src/ethereum.rs
```diff
@@ -1,101 +1,486 @@
...
+
+                let contract_query =
+                    contract.query("privileges", verified.from, None, Options::default(), None);
+                let priviledge_level: U256 = contract_query.wait().unwrap();
```
Ivshti: let's keep this like that for now
but we should change this to an async call ASAP

adapter/src/ethereum.rs
```diff
@@ -1,101 +1,486 @@
...
+        Ok(sess)
+    }
+
+    fn get_auth(&mut self, validator: &ValidatorDesc) -> AdapterResult<String> {
+        match (self.tokens_for_auth.get(&validator.id), &self.wallet) {
+            (Some(token), Some(_)) => Ok(token.to_owned()),
+            (None, Some(wallet)) => {
+                let payload = Payload {
+                    id: validator.id.clone(),
+                    era: usize::try_from(Utc::now().timestamp()).expect("failed to parse utc now"),
```
Ivshti: this is an actual OK case for `expect`! cause there's no realistic way it can fail

adapter/src/ethereum.rs
```diff
@@ -1,101 +1,486 @@
...
+        eth_adapter.unlock().expect("should unlock eth adapter");
+
+        let whoami = eth_adapter.whoami().expect("failed to get whoami");
+        assert_eq!(
+            whoami, "0x2bDeAFAE53940669DaA6F519373f686c1f3d3393",
+            "failed to get correct whoami"
+        );
+
+        // Sign
+        let expected_response =
```
Ivshti: those match with the JS impl, right?

adapter/src/dummy.rs
elpiel: https://github.com/rust-lang/rust/issues/52047

adapter/src/ethereum.rs
samparsky: yes it does

adapter/contract/AdExCore.json
samparsky: I have included it

adapter/src/dummy.rs
```diff
@@ -21,33 +21,40 @@ impl ChannelValidator for DummyAdapter {}
...
+        })
     }
 
     fn unlock(&self) -> AdapterResult<bool> {
         Ok(true)
     }
 
-    fn whoami(&self) -> String {
-        self.identity.to_string()
+    fn whoami(&self) -> AdapterResult<String> {
```
elpiel: Can you document somewhere, what is the use case when you don't have an identity for the adapter?

adapter/src/ethereum.rs
```diff
@@ -1,101 +1,505 @@
...
+                ))
+            }
         };
 
-        Ok(is_same)
+        let address = Address::from_slice(&decoded_adress);
+        let signature = Signature::from_electrum(&decoded_signature);
+        let message = Message::from_slice(&hash_message(state_root));
+
+        match verify_address(&address, &signature, &message) {
```
elpiel: Note for next time: you could use `map_err`, as you are not changing the Ok result, but only the error one.

adapter/src/ethereum.rs
```diff
@@ -1,101 +1,505 @@
...
+        }
+
         Ok(true)
     }
 
-    fn session_from_token(&self, _token: &str) -> AdapterResult<String> {
-        // @TODO
-        Ok("hello".to_string())
+    fn session_from_token(&self, token: &str) -> AdapterResult<Session> {
+        let token_id = token.to_owned()[token.len() - 16..].to_string();
```
elpiel: This can panic, if the token passed doesn't fit in the range: `[token.len() - 16..].` (i.e. token is less than 16 chars)
IMO this should panic it should return an `Err` in that case.

adapter/src/ethereum.rs
```diff
@@ -1,101 +1,505 @@
...
+                }
+                (_, _, _) => {
+                    return Err(AdapterError::Failed(format!(
+                        "{} token string is incorrect",
+                        token
+                    )))
+                }
+            };
+
+        let verified = match ewt_verify(header_encoded, payload_encoded, token_encoded) {
```
elpiel: Note: `map_err` can be used.

adapter/src/ethereum.rs
samparsky: I can't use map_err here because the `Err` arm isn't returning an error but `Ok(false)`

adapter/src/dummy.rs
samparsky: I don't think it's possible to not have an identity for the adapter.
The reason behind the AdapterResult is the because of the Ethereum adapter where you can't call `whoami()` without unlocking the adapter first

adapter/src/ethereum.rs
elpiel: yes, my bad.

adapter/src/ethereum.rs
samparsky: yeah, fixed

adapter/src/ethereum.rs
samparsky: yeah, fixed

primitives/src/util/tests/prep_db.rs
```diff
@@ -5,30 +5,30 @@ use lazy_static::lazy_static;
...
+        ids.insert("publisher".into(), "b7d3f81e857692d13e9d63b232a90f4a1793189e".into());
+        ids.insert("publisher2".into(), "myAwesomePublisher2".into());
+        ids.insert("creator".into(), "awesomeCreator".into());
+        ids.insert("tester".into(), "2892f6C41E0718eeeDd49D98D648C789668cA67d".into());
 
         ids
     };
 
-    pub static ref AUTH: HashMap<&'static str, String> = {
+    pub static ref AUTH: HashMap<String, String> = {
```
elpiel: I've changed that as using `String` will make us type just a bit more when trying to pick a specific key:
`AUTH["my_key".to_string()]` or `AUTH["my_key".to_owned()]` instead of just doing `AUTH["my_key"]`

validator_worker/src/lib.rs
```diff
@@ -52,19 +52,19 @@ mod test {
     use adapter::DummyAdapter;
     use primitives::adapter::AdapterOptions;
     use primitives::config::configuration;
-    use primitives::util::tests::prep_db::{DUMMY_CHANNEL, IDS};
+    use primitives::util::tests::prep_db::{AUTH, DUMMY_CHANNEL, IDS};
     use primitives::{BalancesMap, Channel};
 
     fn setup_iface(channel: &Channel) -> SentryApi<DummyAdapter> {
         let adapter_options = AdapterOptions {
             dummy_identity: Some(IDS["leader"].clone()),
```
elpiel: you did change the `IDS` to also have `String` for key and I am not sure how this is working tbh ?! :thinking:

adapter/src/ethereum.rs
```diff
@@ -1,101 +1,510 @@
...
-        //     )),
-        // };
-        Ok("auth".to_string())
+    fn get_auth(&self, validator: &ValidatorDesc) -> AdapterResult<String> {
+        let mut authorization_tokens = self.authorization_tokens.borrow_mut();
+        match (
+            self.wallet.borrow().clone(),
+            authorization_tokens.get(&validator.id),
+        ) {
+            (Some(_), Some(token)) => Ok(token.to_owned()),
```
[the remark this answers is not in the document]
samparsky: Yeah wallet shouldn't be None

adapter/src/ethereum.rs
```diff
@@ -1,101 +1,510 @@
...
+    password: &Password,
+    payload: &Payload,
+) -> Result<String, Box<dyn Error>> {
+    let header = Header {
+        header_type: "JWT".to_string(),
+        alg: "ETH".to_string(),
+    };
+
+    let header_encoded = base64::encode_config(
+        &serde_json::to_string(&header)?.as_bytes(),
```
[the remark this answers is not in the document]
samparsky: it needs `&` but `as_bytes()` is optional

adapter/src/ethereum.rs
```diff
@@ -1,101 +1,510 @@
...
+    }
+
+    #[test]
+    fn should_generate_correct_ewt_sign_and_verify() {
+        let eth_adapter = setup_eth_adapter();
+        eth_adapter.unlock().expect("should unlock eth adapter");
+
+        let payload = Payload {
+            id: "awesomeValidator".to_string(),
+            era: 10_0000,
```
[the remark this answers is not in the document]
samparsky: yeah

samparsky:
@elpiel The suggested changes have been made

adapter/src/dummy.rs
```diff
@@ -63,27 +70,45 @@ impl Adapter for DummyAdapter {
...
+        }
     }
 
-    fn session_from_token(&self, _token: &str) -> AdapterResult<String> {
-        // @TODO
-        Ok("hello".to_string())
+    fn session_from_token(&self, token: &str) -> AdapterResult<Session> {
+        let identity = self
+            .authorization_tokens
+            .clone()
```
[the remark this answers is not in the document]
samparsky: okay, fixed

adapter/src/lib.rs
```diff
@@ -64,28 +67,55 @@ pub struct EthereumChannel {
     pub creator: String,
     pub token_addr: String,
     pub token_amount: String,
-    pub valid_until: String,
+    pub valid_until: i64,
```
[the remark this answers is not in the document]
samparsky: `i64` works fine but the param to but the param to `new` requires `Datetime`

adapter/src/dummy.rs
```diff
@@ -63,27 +70,45 @@ impl Adapter for DummyAdapter {
...
-        // let future = match participant {
-        //     Some((_, participant)) => ok(participant.token.to_string()),
-        //     None => err(AdapterError::Authentication(
-        //         "Identity not found".to_string(),
-        //     )),
-        // };
-        Ok("auth".to_string())
+        let who = self
+            .session_tokens
+            .clone()
```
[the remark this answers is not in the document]
samparsky: fixed

adapter/src/ethereum.rs
```diff
@@ -1,101 +1,510 @@
...
+                .map_err(|_| map_error("Invalid keystore json provided"))?,
+            None,
+            &Some(self.keystore_pwd.clone()),
+        )
+        .map_err(|_| map_error("Failed to create account"))?;
+
+        self.wallet.replace(Some(account));
+
+        // wallet has been unlocked
         Ok(true)
```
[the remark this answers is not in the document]
samparsky: don't think there is another case for future proofing

adapter/src/ethereum.rs
```diff
@@ -1,101 +1,510 @@
...
+        }
+
+        if check_validator_id_checksum(channel) {
+            return Err(AdapterError::Configuration(
+                "channel.validators: all addresses are checksummed".to_string(),
+            ));
+        }
+        // check if channel is valid
+        let is_channel_valid = EthereumAdapter::is_channel_valid(&self.config, channel);
+        if is_channel_valid.is_err() {
```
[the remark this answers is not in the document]
samparsky: yeah much cleaner

adapter/src/ethereum.rs
```diff
@@ -1,101 +1,510 @@
...
+        let contract_address = Address::from_slice(self.config.ethereum_core_address.as_bytes());
+        let contract = get_contract(&self.config, contract_address, &ADEXCORE_ABI)
+            .map_err(|_| map_error("failed to init core contract"))?;
+
+        let channel_status: U256 = contract
+            .query("states", channel_id, None, Options::default(), None)
+            .wait()
+            .map_err(|_| map_error("contract channel status query failed"))?;
+
+        if channel_status != 1.into() {
```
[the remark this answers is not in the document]
samparsky: yeah changed it to a lazy_static constant to be more descriptive

primitives/src/util/tests/prep_db.rs
samparsky: hmnn .. I am not sure this is correct 
https://doc.rust-lang.org/std/collections/struct.HashMap.html#examples
```
// Check for a specific one.
// When collections store owned values (String), they can still be
// queried using references (&str).
```

adapter/src/lib.rs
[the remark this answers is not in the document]
elpiel: Yeah, I mean it's more about documenting it, creating the EthereumChannel requires Datetime, but when you use the public field `valid_until` from an object, you don't know where it comes from and what it is.

adapter/src/ethereum.rs
```diff
@@ -1,101 +1,492 @@
...
+}
+
+fn get_contract(
+    config: &Config,
+    contract_address: Address,
+    abi: &[u8],
+) -> Result<Contract<Http>, Box<dyn Error>> {
+    let (_eloop, transport) = web3::transports::Http::new(&config.ethereum_network)?;
+    let web3 = web3::Web3::new(transport);
+    let contract = Contract::from_json(web3.eth(), contract_address, abi)?;
```
elpiel: side note: those 2 lines can be simplified, since you already have a result and you don't need `?`
 and then `Ok(...)`, so this:
```
let contract = Contract::from_json(web3.eth(), contract_address, abi)?;
Ok(contract)
```

becomes:
`Contract::from_json(web3.eth(), contract_address, abi)`

primitives/src/util/tests/prep_db.rs
elpiel: Haha, yeah, nice, hooray for rustlang again!

adapter/src/ethereum.rs
samparsky: This is incorrect 

one is of type
```
Result<_, std::boxed::Box<(dyn std::error::Error + 'static)>>
```
the other is of type
```
 std::result::Result<_, ethabi::errors::Error>
```
? boxes the error

samparsky:
@elpiel I think this code is ready to be merged

adapter/src/ethereum.rs
elpiel: Ahh cool, didn't know that, I thought it will automatically do that even without `?`

validator_worker/src/sentry_interface.rs
```diff
@@ -137,30 +154,39 @@ impl<T: Adapter + 'static> SentryApi<T> {
...
             .spec
             .validators
             .into_iter()
             .find(|&v| v.id == whoami);
-        let auth_token = self
-            .adapter
-            .get_auth(validator.unwrap())
-            .expect("Failed to get user auth token");
+
+        let auth_token = adapter.get_auth(validator.unwrap());
```
elpiel: Even though we have fairly small amount of code, I see the confusion that leads to duplicating code by accident.

Here is what I don't understand: Why do we even bother to do this check, if we already have it in the initialization of the SentryInterface?
Why search for the `adapter.whoami()` in the `channle.spec.validators` and then even `validator.unwrap()` the value?
This whole thing can be simplified to just:
```
let auth_token = adapter.get_auth(adapter.whoami());
```
If you follow the implementation, the `adapter.whoami()` will always be included in the Channel validators, but it's just not obvious.

validator_worker/src/sentry_interface.rs
```diff
@@ -137,30 +154,39 @@ impl<T: Adapter + 'static> SentryApi<T> {
...
+        if let Err(e) = auth_token {
```
elpiel: You could map this error and use `?` and get rid off the `unwrap()` bellow.

Same way you did here: https://github.com/AdExNetwork/adex-validator-stack-rust/pull/143/files/e4739e44f81d4c040f627043aa33bbb160c21a71..a781eb41a68ffa5cd31c3b42bfb2de638f773d29#diff-ae6ae3e6a4852afdf6f5eda167faccd7R189

validator_worker/src/sentry_interface.rs
samparsky: `adapter.whoami()` returns a `string`,
`.find(|&v| v.id == whoami);` returns a `ValidatorDesc`

samparsky: Also, the idea behind the unwrap() is because we already check it in `new()` so it would always be there

PR #144, opened 2019-09-19
Issue #18 `.await` syntax

### LINKED ISSUE #18: `.await` syntax
Tracking issue for `.await` syntax and when we are going to be able to update to it.

**Current status:**
It's not possible to move to latest nightly as `tower-web`'s feature `async-await-preview` fails to build when we switch to latest nightly.

```
error[E0432]: unresolved import `std::await`
  --> /home/user/.cargo/registry/src/github.com-1ecc6299db9ec823/tokio-async-await-0.1.7/src/lib.rs:35:9
   |
35 | pub use std::await as std_await;
   |         ^^^^^^^^^^^^^^^^^^^^^^^ no `await` in the root

error: aborting due to previous error
```

elpiel:
It's branched out of #140 and resolves #18


Ivshti:
This looks good to me (although I need to review it again, more carefully); we should merge ASAP

adapter/src/lib.rs
```diff
@@ -64,28 +65,55 @@ pub struct EthereumChannel {
     pub creator: String,
     pub token_addr: String,
     pub token_amount: String,
-    pub valid_until: String,
+    pub valid_until: i64,
     pub validators: String,
     pub spec: String,
 }
 
+impl From<&Channel> for EthereumChannel {
```
Ivshti: this should be `TryFrom` and should handle errors like serialization failures gracefully

think about it - if it doesn't, then it's a DoS attack vector because anyone can pass an invalid spec and make the node crash

PR #154, opened 2019-10-10
Reworking sentry

### LINKED ISSUE #150: Main process for handling incomming requests
As we discussed we will be using [hyper](https://docs.rs/hyper/0.13.0-alpha.4/hyper/)
We might use the probably 0.13 alpha version for the async/await , see [here](https://seanmonstar.com/post/187493499882/hyper-alpha-supports-asyncawait) for more details

elpiel:
Resolves #150 and impl the logging for later use in #156


adapter/src/dummy.rs
```diff
@@ -18,31 +18,18 @@ pub struct DummyAdapter {
...
-                return Err(AdapterError::Configuration(
-                    "dummy_identity, dummy_auth, dummy_auth_tokens required".to_string(),
-                ))
-            }
-        };
-
-        Ok(Self {
-            identity,
+impl DummyAdapter {
+    pub fn init(opts: DummyAdapterOptions, config: &Config) -> Self {
```
elpiel: `DummyAdapter::init` has no failing case, if we pass it's options directly, no need to return a `Result`

primitives/src/adapter.rs
```diff
@@ -33,14 +33,10 @@ impl fmt::Display for AdapterError {
     }
 }
 
-#[derive(Debug, Clone)]
-pub enum AdapterOptions {
```
elpiel: We actauly don't need this, see the `main.rs` of the `validator_worker` and `sentry`. We instantiate them separately, this is not part of the Behaviour of the struct (i.e. behaviour is what Traits should represent, like interfaces in classical OOP).

primitives/src/adapter.rs
```diff
@@ -33,14 +33,10 @@ impl fmt::Display for AdapterError {
...
-    DummAdapter {
-        dummy_identity: String,
-        dummy_auth: HashMap<String, String>,
-        dummy_auth_tokens: HashMap<String, String>,
-    },
-    EthereumAdapter(KeystoreOptions),
+pub struct DummyAdapterOptions {
```
elpiel: This could be moved to the `DummyAdapter` impl., as it's only needed there, for that specific implementation.

primitives/src/adapter.rs
```diff
@@ -56,11 +52,6 @@ pub struct Session {
 }
 
 pub trait Adapter: ChannelValidator + Clone + Debug {
-    type Output;
```
elpiel: The reason for this not being very good idea is that you can't use something like:

let adapter: Box<dyn Adapter> = match predicate {
 .... => one adapter instantiation
 .... => another adapter instantiation
}

and this is because you need to specify an Output, i.e. Box<dyn Adapter<Output=????>> and you can't write any output, because you have different ones in the `match` arms.

sentry/src/lib.rs
```diff
@@ -1,22 +1,101 @@
...
+            port,
+        }
+    }
+
+    pub async fn run(&self) {
+        let addr = ([127, 0, 0, 1], self.port).into();
+
+        dbg!(self.clustered);
+
+        let make_service = make_service_fn(move |_| {
```
elpiel: we only have 1 instance of the Application with a specified Adapter, the magic happens here, were it does create a service and this service is handling incoming connections, so it's cloning the `Adapter` & `Config` for each connection.

primitives/src/adapter.rs
samparsky: I don't think this is unique to this implementation alone

elpiel: You can find more about the associated type error here: https://doc.rust-lang.org/nightly/error-index.html#E0191
if you want to specify `Box<dyn Adapter>` you need to specify the associated type as well.

sentry/src/main.rs
```diff
@@ -1,104 +1,117 @@
...
+                .await
+        }
+    }
 }
 
-impl TryFrom<Vars> for Config {
-    type Error = domain::DomainError;
+fn logger() -> Logger {
+    let decorator = TermDecorator::new().build();
+    let drain = PrefixedCompactFormat::new("sentry", decorator).fuse();
```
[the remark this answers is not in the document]
elpiel: we can use the Key-value pairs:
`info!(&self.logger, "Listening on port {}!", self.port; o!("module" => "Sentry lib"));`

`info!(&self.logger, "Listening on port {}!", self.port; "module" => "Sentry lib");`

elpiel:
ping @samparsky can you check the merge of `dev` and approve the pr?

PR #155, opened 2019-10-11
fix: validator worker startup

### LINKED ISSUE #146: Refactor: Implement validator worker startup process

samparsky:
Fixes #146

### DIFF
--- a/primitives/src/sentry.rs
+++ b/primitives/src/sentry.rs
@@ -89,7 +89,7 @@ pub struct ValidatorMessage {
 #[derive(Serialize, Deserialize, Debug)]
 #[serde(rename_all = "camelCase")]
 pub enum ValidatorMessageResponse {
-    ValidatorMessages(Option<Vec<ValidatorMessage>>),
+    ValidatorMessages(Vec<ValidatorMessage>),
 }
 
 #[derive(Serialize, Deserialize, Debug)]
--- a/primitives/src/sentry.rs
+++ b/primitives/src/sentry.rs
@@ -80,10 +80,16 @@ pub struct SuccessResponse {
 }
 
 #[derive(Serialize, Deserialize, Debug)]
-pub struct ValidatorMessageResponse {
+pub struct ValidatorMessage {
     pub from: String,
     pub received: DateTime<Utc>,
-    pub msg: Vec<MessageTypes>,
+    pub msg: MessageTypes,
+}
+
+#[derive(Serialize, Deserialize, Debug)]
+#[serde(rename_all = "camelCase")]
+pub enum ValidatorMessageResponse {
+    ValidatorMessages(Option<Vec<ValidatorMessage>>),
 }
 
 #[derive(Serialize, Deserialize, Debug)]
--- a/validator_worker/src/follower.rs
+++ b/validator_worker/src/follower.rs
@@ -1,5 +1,6 @@
 use std::error::Error;
 
+use futures::compat::Future01CompatExt;
 use primitives::adapter::Adapter;
 use primitives::validator::{ApproveState, MessageTypes, NewState, RejectState};
 use primitives::BalancesMap;
@@ -30,7 +31,7 @@ pub async fn tick<A: Adapter + 'static>(iface: &SentryApi<A>) -> Result<(), Box<
     };
 
     let our_latest_msg_response = iface
-        .get_our_latest_msg(&["ApproveState","RejectState"])
+        .get_our_latest_msg(&["ApproveState", "RejectState"])
         .await?;
 
     let our_latest_msg_state_root = match our_latest_msg_response {
@@ -65,7 +66,12 @@ async fn on_new_state<'a, A: Adapter + 'static>(
         return Ok(on_error(&iface, &new_state, InvalidNewState::RootHash).await);
     }
 
-    let adapter = iface.adapter.lock().await;
+    let adapter = iface
+        .adapter
+        .read()
+        .compat()
+        .await
+        .expect("on_new_state: failed to acquire read lock adapter");
 
     if !adapter.verify(
         &iface.channel.spec.validators.leader().id,
--- a/validator_worker/src/follower.rs
+++ b/validator_worker/src/follower.rs
@@ -23,14 +23,14 @@ enum NewStateResult {
 
 pub async fn tick<A: Adapter + 'static>(iface: &SentryApi<A>) -> Result<(), Box<dyn Error>> {
     let from = iface.channel.spec.validators.leader().id.clone();
-    let new_msg_response = iface.get_latest_msg(from, "NewState".to_string()).await?;
+    let new_msg_response = iface.get_latest_msg(from, &["NewState"]).await?;
     let new_msg = match new_msg_response {
         Some(MessageTypes::NewState(new_state)) => Some(new_state),
         _ => None,
     };
 
     let our_latest_msg_response = iface
-        .get_our_latest_msg("ApproveState+RejectState".to_string())
+        .get_our_latest_msg(&["ApproveState","RejectState"])
         .await?;
 
     let our_latest_msg_state_root = match our_latest_msg_response {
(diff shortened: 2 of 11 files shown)


validator_worker/src/main.rs
```diff
@@ -98,26 +120,100 @@ fn main() {
...
+
+async fn infinite<A: Adapter + 'static>(args: Args<A>) -> Result<(), ()> {
+    loop {
+        let arg = args.clone();
+        let delay_future =
+            Delay::new(Instant::now().add(Duration::from_secs(arg.config.wait_time as u64)));
+        let joined = join(iterate_channels(arg), delay_future.compat());
+        match joined.await {
+            (_, Err(e)) => eprintln!("{}", e),
+            _ => println!("finished processing channels"),
```
elpiel: Is this going to show up every `arg.config.wait_time` seconds? Is this necessary for the output?

validator_worker/src/main.rs
```diff
@@ -98,26 +120,100 @@ fn main() {
...
+        };
+    }
+}
+
+async fn iterate_channels<A: Adapter + 'static>(args: Args<A>) -> Result<(), ()> {
+    let channels = all_channels(&args.sentry_url, args.whoami.clone())
+        .await
+        .expect("Failed to get channels");
```
elpiel: shouldn't we then just return from the function? I think if Sentry is down for a bit, then the validator will exit this way, which is not idea IMO.

validator_worker/src/main.rs
```diff
@@ -98,26 +120,100 @@ fn main() {
...
+    let channels_size = channels.len();
+
+    try_join_all(
+        channels.into_iter().map(|channel| {
+            validator_tick(args.adapter.clone(), channel, &args.config, &args.whoami)
+        }),
+    )
+    .await
+    .expect("Failed to iterate channels");
```
elpiel: You won't need this if the `validator_tick` is not returning anything ( see comment bellow).

validator_worker/src/main.rs
```diff
@@ -98,26 +120,100 @@ fn main() {
...
+
+    Ok(())
+}
+
+async fn validator_tick<A: Adapter + 'static>(
+    adapter: Arc<Mutex<A>>,
+    channel: Channel,
+    config: &Config,
+    whoami: &str,
+) -> Result<(), ()> {
```
elpiel: I think you can just remove the return type and not use a Result.

validator_worker/src/sentry_interface.rs
```diff
@@ -114,13 +108,24 @@ impl<T: Adapter + 'static> SentryApi<T> {
             .and_then(|mut res: Response| res.json::<ValidatorMessageResponse>())
             .compat();
 
-        future.await
+        match future.await {
```
elpiel: Side note: We can simplify this.

validator_worker/src/sentry_interface.rs
```diff
@@ -225,23 +227,19 @@ fn handle_http_error(e: reqwest::Error, url: &str) {
...
     if first_page.total_pages < 2 {
         Ok(first_page.channels)
     } else {
         let mut all: Vec<ChannelAllResponse> = try_join_all(
-            (0..first_page.total_pages).map(|i| fetch_page(url.clone(), i, validator.clone())),
+            (0..first_page.total_pages).map(|i| fetch_page(url.clone(), i, whoami.clone())),
         )
         .await
-        .unwrap();
+        .expect("Failed to fetch all channels");
```
elpiel: Shouldn't this be returned as an error, instead of crashing the validator? Again - Sentry down for a bit, network issues and etc., will cause it to blow up and stop the validator.

primitives/src/sentry.rs
```diff
@@ -80,10 +80,16 @@ pub struct SuccessResponse {
...
+pub struct ValidatorMessage {
     pub from: String,
     pub received: DateTime<Utc>,
-    pub msg: Vec<MessageTypes>,
+    pub msg: MessageTypes,
+}
+
+#[derive(Serialize, Deserialize, Debug)]
+#[serde(rename_all = "camelCase")]
+pub enum ValidatorMessageResponse {
```
elpiel: I think this should be in sentry_interface and not public as it's only used for deserializing in `sentry_interface`

samparsky: it's needed it in sentry too for `ValidatorMessageResponse`

validator_worker/src/main.rs
samparsky: That's added because of `try_join_all`

validator_worker/src/sentry_interface.rs
samparsky: yeah true

validator_worker/src/sentry_interface.rs
samparsky: This could be helpful in cases of wrong sentry url

validator_worker/src/main.rs
samparsky: It's meant to be a performance metric helper. The println! will be changed to logger when we have the implementation.

validator_worker/src/main.rs
samparsky: yeah true

validator_worker/src/main.rs
samparsky: Also, clippy shows warnings if the error variant isn't handled

validator_worker/src/main.rs
samparsky: If you don't handle Err variant for Result<>, clippy would show warnings

primitives/src/sentry.rs
```diff
@@ -80,10 +80,16 @@ pub struct SuccessResponse {
...
+    ValidatorMessages(Option<Vec<ValidatorMessage>>),
```
Ivshti: do we need an `Option` here; what extra meaning does it add on top of just sending back an empty vector?

validator_worker/src/follower.rs
```diff
@@ -24,24 +24,20 @@ enum NewStateResult {
...
-            MessageTypes::NewState(new_state) => Some(new_state.clone()),
-            _ => None,
-        });
+    let new_msg = match new_msg_response {
+        Some(MessageTypes::NewState(new_state)) => Some(new_state),
+        _ => None,
+    };
+
     let our_latest_msg_response = iface
         .get_our_latest_msg("ApproveState+RejectState".to_string())
```
Ivshti: perhaps this api would be nicer if it gets `&[&str]` so that we avoid the `+` hack

validator_worker/src/lib.rs
```diff
@@ -61,12 +61,14 @@ mod test {
         };
         let config = configuration("development", None).expect("Dev config should be available");
         let dummy_adapter = DummyAdapter::init(adapter_options, &config).expect("init adadpter");
+        let whoami = dummy_adapter.whoami();
 
-        SentryApi::new(
-            Arc::new(RwLock::new(dummy_adapter)),
+        SentryApi::init(
+            Arc::new(Mutex::new(dummy_adapter)),
```
Ivshti: is there a reason for using a `Mutex` rather than a `Rwlock`

validator_worker/src/main.rs
elpiel: You can still do a `_ => {}`, but that's ok for now. If it's annoying posting `finished processing channels` every `wait_time` seconds we can remove it.

validator_worker/src/sentry_interface.rs
```diff
@@ -227,21 +240,19 @@ fn handle_http_error(e: reqwest::Error, url: &str) {
...
-        .await
-        .expect("Failed to get channels from sentry url");
+    let first_page = fetch_page(url.clone(), 0, whoami.clone()).await?;
+
     if first_page.total_pages < 2 {
         Ok(first_page.channels)
     } else {
         let mut all: Vec<ChannelAllResponse> = try_join_all(
-            (0..first_page.total_pages).map(|i| fetch_page(url.clone(), i, validator.clone())),
+            (0..first_page.total_pages).map(|i| fetch_page(url.clone(), i, whoami.clone())),
```
elpiel: Why are we fetching the first page again btw?

Issue #160, opened 2019-10-24
Refetching first page in `SentryApi::all_channels`


elpiel:
Follow up on https://github.com/AdExNetwork/adex-validator-stack-rust/pull/155/files#r338465394

PR #161, opened 2019-10-24
Issue #158 fix validator worker

### LINKED ISSUE #158: Fix: validator_worker bugs
- [x] Fix sentry ValidatorMessageResponse, EventAggregateResponse 
- [x] getBalanceLeaf incorrect encoding
- [x] getStateRootHash incorrect encoding
- [x] Fix RwLock follower Write deadlock
- [x] Fix new states should not be produced when there are no new aggregates (producer tick)
- [x] Fix is_channel_not_exhausted

samparsky:
Fixes #158  #160 
Also includes the ValidatorId type


adapter/src/dummy.rs
```diff
@@ -34,24 +34,29 @@ impl Adapter for DummyAdapter {
         Ok(())
     }
 
-    fn whoami(&self) -> String {
+    fn whoami(&self) -> ValidatorId {
```
elpiel: Why not a reference and wherever we need owned, we can `clone()` or `to_owned()`?

adapter/src/dummy.rs
```diff
@@ -34,24 +34,29 @@ impl Adapter for DummyAdapter {
...
         self.identity.clone()
     }
 
     fn sign(&self, state_root: &str) -> AdapterResult<String> {
         let signature = format!(
             "Dummy adapter signature for {} by {}",
             state_root,
-            self.whoami()
+            self.whoami().to_string()
```
elpiel: No need for `.to_string()` when `Display` is implemented

adapter/src/dummy.rs
```diff
@@ -34,24 +34,29 @@ impl Adapter for DummyAdapter {
...
+        &self,
+        signer: &ValidatorId,
+        _state_root: &str,
+        signature: &str,
+    ) -> AdapterResult<bool> {
         // select the `identity` and compare it to the signer
         // for empty string this will return array with 1 element - an empty string `[""]`
         let is_same = match signature.rsplit(' ').take(1).next() {
-            Some(from) => from == signer,
+            Some(from) => from == signer.to_string(),
```
elpiel: or `Some(ref from) => from == signer`

adapter/src/ethereum.rs
```diff
@@ -66,8 +66,10 @@ impl EthereumAdapter {
             }
         };
 
+        let identity = ValidatorId::try_from(address.as_str())?;
```
elpiel: `address(owned ValidatorId) -> &str -> ValidatorId::try_from` Why?! Just use the address, it's an owned value.

adapter/src/ethereum.rs
```diff
@@ -113,29 +115,17 @@ impl Adapter for EthereumAdapter {
...
-            (Ok(address), Ok(sig)) => (address, sig),
-            _ => {
-                return Err(AdapterError::Signature(
-                    "invalid signature or address".to_string(),
-                ))
-            }
-        };
-
-        let address = Address::from_slice(&decoded_adress);
+    fn verify(&self, signer: &ValidatorId, state_root: &str, sig: &str) -> AdapterResult<bool> {
```
elpiel: we take reference here as well, so yet another reason to return a reference from `whoami`
Also `get_auth` takes a reference.

adapter/src/lib.rs
```diff
@@ -194,7 +195,9 @@ impl EthereumChannel {
         contract_addr: &str,
         balance_root: &str,
     ) -> Result<[u8; 32], Box<dyn Error>> {
-        get_signable_state_root(contract_addr, balance_root)
+        let root = <[u8; 32]>::from_hex(balance_root).unwrap();
```
elpiel: at least `expect`? Although it might be good to handle the errors even if we are sure that `contact_addr` & `balance_root` are ok.

primitives/src/channel.rs
```diff
@@ -103,7 +105,7 @@ impl SpecValidators {
         &self.0[1]
     }
 
-    pub fn find(&self, validator_id: &str) -> SpecValidator<'_> {
+    pub fn find(&self, validator_id: ValidatorId) -> SpecValidator<'_> {
```
elpiel: can take a reference, instead of needing to clone the `ValidatorId`

primitives/src/channel_validator.rs
```diff
@@ -1,11 +1,16 @@
...
 use crate::config::Config;
+use crate::ValidatorId;
 use chrono::Utc;
+use std::convert::TryFrom;
 
 pub trait ChannelValidator {
     fn is_channel_valid(config: &Config, channel: &Channel) -> Result<(), ChannelError> {
         let identity = &config.clone().identity.unwrap_or_else(|| "".to_string());
-        let adapter_channel_validator = match channel.spec.validators.find(identity) {
+        let validator_identity = ValidatorId::try_from(&identity[..]).map_err(|_| {
```
elpiel: I think `&identity` is enough here or is there another reason for this?

primitives/src/channel_validator.rs
```diff
@@ -45,7 +50,9 @@ pub fn all_validators_listed(validators: &SpecValidators, whitelist: &[String])
     } else {
         let found_validators = whitelist
             .iter()
-            .filter(|&allowed| {
+            .map(|identity| ValidatorId::try_from(&identity[..]))
```
elpiel: Shouldn't the `whitelist` also be `ValidatorId` slice?
PS: Also this `&identity[..]` here again.

primitives/src/validator.rs
```diff
@@ -17,14 +19,51 @@ pub enum ValidatorError {
...
+        })?;
+        let mut id: [u8; 20] = [0; 20];
+        id.copy_from_slice(&result[..]);
+        Ok(Self(id))
+    }
+}
+
+impl fmt::Display for ValidatorId {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "{}", format!("0x{}", hex::encode(self.0)))
```
elpiel: You can use the SerHex btw :wink: 
https://github.com/fspmarshall/serde-hex/issues/3#issuecomment-497826234

primitives/src/validator.rs
```diff
@@ -34,7 +73,7 @@ pub struct ValidatorDesc {
 #[derive(Serialize, Deserialize, Debug, Clone)]
 #[serde(rename_all = "camelCase")]
 pub struct Accounting {
-    #[serde(rename = "last_ev_aggr")]
+    #[serde(rename = "lastEvAggr")]
```
elpiel: good catch! :+1:

validator_worker/src/core/fees.rs
```diff
@@ -77,7 +77,7 @@ fn distribute_fee<'a>(
 
         if fee_rounded > 0.into() {
             let entry = balances
-                .entry(validator.id.clone())
+                .entry(validator.id.to_string().clone())
```
elpiel: I think now we don't need the `clone()`

validator_worker/src/follower.rs
```diff
@@ -66,44 +65,47 @@ async fn on_new_state<'a, A: Adapter + 'static>(
...
+        drop(adapter);
         return Ok(on_error(&iface, &new_state, InvalidNewState::Signature).await);
     }
 
     let last_approve_response = iface.get_last_approved().await?;
     let prev_balances = last_approve_response
         .last_approved
         .and_then(|last_approved| last_approved.new_state)
-        .map_or(Default::default(), |new_state| new_state.balances);
+        .map_or(Default::default(), |new_state| match new_state.msg {
```
elpiel: isn't this taking quite a bit out of the expressiveness? I.e. instead of having a `new_state` we have the whole message and we need to always match it.

validator_worker/src/heartbeat.rs
```diff
@@ -69,5 +60,5 @@ pub async fn heartbeat<A: Adapter + 'static>(
 }
 
 fn is_channel_not_exhausted(channel: &Channel, balances: &BalancesMap) -> bool {
```
elpiel: this is not very good, usually it's good to stick to the positive and shorter names, that's why in Rust you don't have `is_not_empty()` for example.

validator_worker/src/main.rs
```diff
@@ -170,33 +178,26 @@ async fn iterate_channels<A: Adapter + 'static>(args: Args<A>) -> Result<(), ()>
...
-    let index = channel
-        .spec
-        .validators
-        .into_iter()
-        .position(|v| v.id == *whoami);
     let duration = Duration::from_secs(config.validator_tick_timeout as u64);
-    match index {
-        Some(0) => {
+
+    match channel.spec.validators.find(whoami.to_owned()) {
```
elpiel: just use a reference, it will make it much cleaner, we don't need owned value for `find`

validator_worker/src/sentry_interface.rs
```diff
@@ -33,22 +34,18 @@ impl<T: Adapter + 'static> SentryApi<T> {
...
-            .spec
-            .validators
-            .into_iter()
-            .find(|&v| v.id == whoami)
-        {
-            Some(v) => {
-                let validator_url = format!("{}/channel/{}", v.url, channel.id);
+        match channel.spec.validators.find(whoami.clone()) {
+            SpecValidator::Leader(v) | SpecValidator::Follower(v) => {
+                let channel_id = format!("0x{}", hex::encode(&channel.id));
```
elpiel: Again the trick with SerHex can be used :stuck_out_tongue: and one more down bellow.

validator_worker/src/sentry_interface.rs
```diff
@@ -162,47 +151,54 @@ impl<T: Adapter + 'static> SentryApi<T> {
...
             .send()
-            .and_then(|mut res: Response| res.json::<EventAggregateResponse>())
-            .map_err(|e| Box::new(ValidatorWorker::Failed(e.to_string())));
+            .map_err(|e| Box::new(ValidatorWorker::Failed(e.to_string())))
+            .compat()
+            .await?
+            .json()
+            .map_err(|e| Box::new(ValidatorWorker::Failed(e.to_string())))
+            .compat()
+            .await?;
```
elpiel: you can just remove the `;` and the `Ok(result)`?

primitives/src/sentry.rs
```diff
@@ -1,16 +1,17 @@
...
 use serde::{Deserialize, Serialize};
+use serde_hex::{SerHex, StrictPfx};
 use std::collections::HashMap;
 
 #[derive(Serialize, Deserialize, Debug)]
 #[serde(rename_all = "camelCase")]
 pub struct LastApproved {
     /// NewState can be None if the channel is brand new
-    pub new_state: Option<NewState>,
+    pub new_state: Option<ValidatorMessage>,
```
elpiel: This forces us to always check if it's actually a `NewState`, isn't it better to leave the explicit types here?

samparsky: What's returned from sentry isn't `NewState` but a `ValidatorMessage`

validator_worker/src/sentry_interface.rs
samparsky: I felt looks much better to the eyes this way, I didn't want to have `Ok()` with the huge content within it

validator_worker/src/follower.rs
samparsky: Again the response from sentry isn't NewState but a `ValidatorMessage`

validator_worker/src/core/fees.rs
samparsky: yeah true

primitives/src/channel_validator.rs
samparsky: identity is a `String` appending it with `[...]` is a shortcut for converting it into a `str` slice

samparsky: TryFrom is implemented for `&str`

adapter/src/dummy.rs
samparsky: `signer` is a `ValidatorId`, `to_string` converts it to the string representation, so this won't work

validator_worker/src/sentry_interface.rs
elpiel: You can remove the `?` as well, since `.await` will give you already a `Result` no need of `Ok(...)`

primitives/src/sentry.rs
elpiel: but we can move this match where we fetch it, instead of doing it everywhere else we use `LastApproved`

primitives/src/channel_validator.rs
elpiel: Yes, but you have a Deref for String, which means you can pass `&string` and it will convert it to `&str`

adapter/src/ethereum.rs
elpiel: We've discussed this already, forgot to remove it, but you can still use `&address` because of the `Deref` impl of `String`

primitives/src/channel_validator.rs
samparsky: No, you can't, it's not possible. It requires implementing TryFrom for &String

samparsky: You can read more here 
https://stackoverflow.com/questions/48244892/should-rust-implementations-of-from-tryfrom-target-references-or-values

primitives/src/sentry.rs
samparsky: Right now, we only have a single place that we use `LastApproved`, I'd say It's okay right now but if it occurs that we use it at several places with this same requirements then we could abstract it

validator_worker/src/sentry_interface.rs
samparsky: The SerHex crate isn't used in validator_worker

primitives/src/channel_validator.rs
elpiel: ok, then why not just use `.as_str()`, and btw you don't need the `&` on the line above, you have an owned `String` there

primitives/src/sentry.rs
elpiel: It's just being less type safe on the top level as it's not bound to what it is actually meaning. I.e. the field says `new_state`, but the value is `ValidatorMessage`, which might not be a `ValidatorMessage::NewState`

validator_worker/src/sentry_interface.rs
elpiel: ok, cool!

primitives/src/sentry.rs
```diff
@@ -48,15 +63,17 @@ pub struct Earner {
 #[derive(Debug, Serialize, Deserialize)]
 #[serde(rename_all = "camelCase")]
 pub struct EventAggregate {
-    pub channel_id: String,
+    #[serde(with = "SerHex::<StrictPfx>")]
+    pub channel_id: [u8; 32],
```
elpiel: We can make type alias for the `channel_id`, just an improvement we can do.

primitives/src/validator.rs
```diff
@@ -17,14 +19,60 @@ pub enum ValidatorError {
...
+        hex::encode(self.0)
+    }
+    pub fn to_hex_checksummed_string(&self) -> String {
+        eth_checksum::checksum(&format!("0x{}", hex::encode(self.0)))
+    }
+}
+
+impl TryFrom<&str> for ValidatorId {
+    type Error = DomainError;
+    fn try_from(value: &str) -> Result<Self, Self::Error> {
```
elpiel: Now I realised that we might have a problem here. What if a shorter value is passed? It won't fill up all the bytes and it will create a `ValidatorId` instead of error.

validator_worker/src/main.rs
```diff
@@ -134,6 +141,8 @@ fn run<A: Adapter + 'static>(is_single_tick: bool, sentry_url: &str, config: &Co
     } else {
         tokio::run(infinite(args).boxed().compat());
```
elpiel: FYI: you can use the `#[tokio::main]` for the `main()`

primitives/src/validator.rs
samparsky: A shorter value can't be passed

primitives/src/validator.rs
```diff
@@ -17,14 +19,60 @@ pub enum ValidatorError {
...
+            &value[2..]
+        } else {
+            value
+        };
+
+        let result = hex::decode(hex_value).map_err(|_| {
+            DomainError::InvalidArgument("Failed to deserialize validator id".to_string())
+        })?;
+        let mut id: [u8; 20] = [0; 20];
+        id.copy_from_slice(&result[..]);
```
elpiel: `copy_from_slice` will panic if the a slice that's not the same length is passed to it.

primitives/src/validator.rs
```diff
@@ -48,6 +48,14 @@ impl TryFrom<&str> for ValidatorId {
         let result = hex::decode(hex_value).map_err(|_| {
             DomainError::InvalidArgument("Failed to deserialize validator id".to_string())
         })?;
+
+        if result.len() != 20 {
```
elpiel: please, don't do this multiple times. We already have a check above.
`hex_value` can be Result, we can check first for `42` then `40` else do `Err` and at the end just `?`.
This will remove the split that is currently here and make it consistent.

elpiel: Ok, lets leave it like that.

PR #163, opened 2019-10-25
Issue #159 event reducer

### LINKED ISSUE #159: Event Reducer
For now we will implement the logic from `05-stable` branch in the JS implementation:

- [x] reduce #163 
- [x] merge_impression_ev #163 

[Reference JS impl](https://github.com/AdExNetwork/adex-validator/blob/05-stable/services/sentry/lib/eventReducer.js)

elpiel:
Solves #159 
- [X] Implement
- [X] Write tests from the JS impl

**NOTE: Uses unmerged PR from Redis and an `alpha` version for now**


sentry/src/event_reducer.rs
```diff
@@ -0,0 +1,101 @@
...
+        .event_counts
+        .entry(earner.into())
+        .or_insert_with(|| 0.into());
+    *event_counts += &1.into();
+
+    let event_payouts = impression
+        .event_payouts
+        .entry(earner.into())
+        .or_insert_with(|| 0.into());
+    *event_payouts += &channel.spec.min_per_impression;
```
samparsky: There seems to be a condition missing it's 
```
new BN(channel.spec.minPerImpression || 1)
```

https://github.com/AdExNetwork/adex-validator/blob/05-stable/services/sentry/lib/eventReducer.js#L47

elpiel: It's not an `Option`, I am not sure what is the intend of `|| 1` exactly - if there is a value or if it's `0` then use `1`, can you give me details about that?

samparsky: if `minPerImpression` is undefined use 1 the default price

elpiel: We do not use an `Option` in the Spec for the `minPerImpression` so we don't need this check. It will always be `>= 0` as the `Bignum` allows only unsigned numbers

PR #165, opened 2019-10-28
Issue #162 Impl check_access

### LINKED ISSUE #162: Impl checkAccess
[JS reference implementation](https://github.com/AdExNetwork/adex-validator/blob/05-stable/services/sentry/lib/access.js)

elpiel:
Resolves #162


sentry/src/access.rs
```diff
@@ -0,0 +1,147 @@
...
+    let current_time = Utc::now();
+
+    if current_time > channel.valid_until {
+        return Response::ChannelIsExpired;
+    }
+
+    if current_time > channel.spec.withdraw_period_start && !events.iter().all(is_close_event) {
+        return Response::ChannelIsPastWithdrawPeriod;
+    }
+
```
samparsky: There is a missing check for not rate-limiting channel creator

sentry/src/access.rs
```diff
@@ -0,0 +1,147 @@
...
+use crate::Session;
+
+#[derive(Debug, PartialEq, Eq)]
+pub enum Response {
+    OnlyCreatorCanCloseChannel,
+    ChannelIsExpired,
+    ChannelIsPastWithdrawPeriod,
+    RulesError(String),
+    Success,
+}
```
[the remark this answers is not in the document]
elpiel: ReduceResult, Result? `Response` is too vague indeed.
Keep in mind it's also namespaced, so even if we have many `Response` or `Xxxx` structs it doesn't really matter.

sentry/src/access.rs
elpiel: 1 line bellow? Isn't this what you mean?
https://github.com/AdExNetwork/adex-validator/blob/05-stable/services/sentry/lib/access.js#L34

samparsky: It's missing in `05-stable` but a channel creator shouldn't be rate-limited
https://github.com/AdExNetwork/adex-validator/blob/master/services/sentry/lib/access.js#L37

sentry/src/access.rs
```diff
@@ -0,0 +1,147 @@
...
+
+// @TODO: Make pub(crate)
+pub async fn check_access(
+    redis: &SharedConnection,
+    session: &Session,
+    rate_limit: &RateLimit,
+    channel: &Channel,
+    events: &[Event],
+) -> Response {
```
samparsky: `check_access` should return a Result, it provides a better API

sentry/src/access.rs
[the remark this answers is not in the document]
samparsky: In the JS impl we don't use the error types for anything, we mostly care if an error occurs
https://github.com/AdExNetwork/adex-validator/blob/05-stable/services/sentry/lib/access.js#L25

PR #170, opened 2019-11-04
/cfg route

elpiel:
/cfg route related to #9


sentry/src/lib.rs
```diff
@@ -71,8 +82,13 @@ where
     }
 }
 
-async fn handle_routing(req: Request<Body>, adapter: impl Adapter) -> Response<Body> {
-    if req.uri().path().starts_with("/channel") {
+async fn handle_routing(
+    req: Request<Body>,
+    (adapter, config): (impl Adapter, Config),
+) -> Response<Body> {
+    if req.uri().path().starts_with("/cfg") && req.method() == Method::GET {
```
samparsky: Can we use `match` statement here instead of `if else` ?

elpiel: We could try, but I rather leave it as simple as possible right now. we need to do something like:
```
match (req.uri().path(), req.method()) {
   (string, method) if string.starts_with("/cfg") && method == Method::GET => ....,
   (string) if string.starts_with("/channel") => ....
}
```

way too much complications for now + if we decide we need Regex for future route we need to probably revert back to `if-else`s

samparsky: I don't think it's too much complications, it's much better.
Also, thinking about it I don't think we should be using `starts_with` it should be equilvalent string match i.e `path() == "/cfg"` not starts_with.
```
match (req.uri().path(), req.method()) {
   ("/cfg", Method::GET) => ()..,
```
Also on Regex we would be using `RegexSet` https://docs.rs/regex/1.3.1/regex/#example-match-multiple-regular-expressions-simultaneously. And a match statement should fit in okay there too

elpiel: Done.

sentry/src/lib.rs
```diff
@@ -2,13 +2,24 @@
...
     pub mod channel;
+    pub mod cfg {
+        use crate::ResponseError;
+        use hyper::{Body, Response};
+        use primitives::Config;
+
+        pub fn return_config(config: &Config) -> Result<Response<Body>, ResponseError> {
+            let config_str = serde_json::to_string(config)?;
+
+            Ok(Response::builder().body(Body::from(config_str)).unwrap())
```
samparsky: This is missing the `application/json` response content type

sentry/src/lib.rs
```diff
@@ -71,11 +82,16 @@ where
...
-        crate::routes::channel::handle_channel_routes(req, adapter).await
-    } else {
-        Err(ResponseError::NotFound)
+async fn handle_routing(
+    req: Request<Body>,
+    (adapter, config): (impl Adapter, Config),
+) -> Response<Body> {
+    match (req.uri().path(), req.method()) {
+        ("/cfg", &Method::GET) => crate::routes::cfg::return_config(&config),
+        (route, _) if route.starts_with("/channel") => {
```
samparsky: Instead of having routes defined at multiple levels within the application, can we move all route definitions here? We can move the `/channel/list` & `/channel` route definitions here

elpiel: This change request is not part of this PR. Please consider first opening an issue to discuss how to do it and then to make the change instead of blocking this issue.

sentry/src/lib.rs
elpiel: Good catch!

PR #172, opened 2019-11-05
SpecValidators tuple instead of array

elpiel:
- Switch to a tuple in the struct to hold the (Leader, Follower) pair
- Change the `impl From`  from array to tuple


elpiel:
Apparently I've ran `cargo make` inside `sentry` crate, instead of the whole project. We still use the `From` impl inside tests. Now it should be good to go again.

PR #179, opened 2019-11-06
Fix get auth

### LINKED ISSUE #174: EthereumAdapter mutability
Remove both caches for `get_auth` & `session_from_token` for EthereumAdapter as:
- [X] `get_auth` - has a  minor micro optimization, primarily for the JS and the computation needed to be done there - PR #179 
- [X] `session_from_token` #176 - it calls the contact, we do need to cache it, but the caching can be done on application level with `redis` - PR #178 

For `unlock()`, we should remain mutable. What we will do for the `validator_worker` is to have a mutable Adapter when we start the validator, and before starting the async process, we should `unlock()` it. After words, we don't need a mutable Adapter any more.

### LINKED ISSUE #180: Adapter Bug: Adapter is not unlocked
Adapter should be unlocked

samparsky:
Fixes #174 `get_auth`
* Removes RwLock for Adapter
* Fixes #180 `unlock` adapter bug

### DIFF
--- a/adapter/src/ethereum.rs
+++ b/adapter/src/ethereum.rs
@@ -226,7 +226,7 @@ impl Adapter for EthereumAdapter {
     fn get_auth(&self, validator: &ValidatorId) -> AdapterResult<String> {
         let wallet = self
             .wallet
-            .clone()
+            .as_ref()
             .ok_or_else(|| AdapterError::Configuration("failed to unlock wallet".to_string()))?;
 
         let era = Utc::now().timestamp_millis() as f64 / 60000.0;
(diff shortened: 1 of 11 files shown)


adapter/src/ethereum.rs
```diff
@@ -227,33 +223,22 @@ impl Adapter for EthereumAdapter {
...
-            }
-            (_, _) => Err(AdapterError::Configuration(
-                "failed to unlock wallet".to_string(),
-            )),
-        }
+    fn get_auth(&self, validator: &ValidatorId) -> AdapterResult<String> {
+        let wallet = self
+            .wallet
+            .as_ref()
+            .ok_or_else(|| AdapterError::Configuration("failed to unlock wallet".to_string()))?;
```
elpiel: `Wallet is not unlocked` is more appropriate message IMO as we are not actually unlocking the wallet.

primitives/src/adapter.rs
```diff
@@ -59,7 +59,7 @@ pub struct Session {
     pub uid: ValidatorId,
 }
 
-pub trait Adapter: ChannelValidator + Send + Clone + Debug {
+pub trait Adapter: ChannelValidator + Send + Sync + Clone + Debug {
```
elpiel: Do we need `Sync` or should we just clone the `Adapter` instead of passing a reference to it?

PR #181, opened 2019-11-07
Issue #177 auth middleware

### LINKED ISSUE #177: Auth middleware
- implement the middleware https://github.com/AdExNetwork/adex-validator/blob/05-stable/middlewares/auth.js
- implement an Application cache

### LINKED ISSUE #182: Cors middleware

elpiel:
Resolves #177 
Resolves #182


sentry/src/lib.rs
```diff
@@ -123,5 +165,5 @@ pub fn bad_request(error: Box<dyn std::error::Error>) -> Response<Body> {
 pub struct Session {
     pub era: i64,
     pub uid: String,
-    pub ip: String,
+    pub ip: Option<String>,
```
elpiel: Accordingly to the JS impl [reference](https://github.com/AdExNetwork/adex-validator/blob/05-stable/routes/channel.js#L258)

sentry/src/middleware/auth.rs
```diff
@@ -0,0 +1,95 @@
...
+                    }
+                })
+                .transpose()
+        })
+        .transpose()?;
+
+    if let Some(ref token) = token {
+        let adapter_session = match redis::cmd("GET")
+            .arg(token)
+            .query_async::<_, Option<String>>(&mut redis.clone())
```
elpiel: AFAIK there is no deserializatioin/serialization for Redis yet, only this (serialization only):
https://github.com/OneSignal/serde-redis

sentry/src/middleware/auth.rs
```diff
@@ -0,0 +1,95 @@
...
+            .await?
+        {
+            Some(session) => {
+                match serde_json::from_str::<AdapterSession>(&session) {
+                    Ok(session) => session,
+                    Err(serde_error) => {
+                        // On a Deserialization error remove the key from Redis as a precaution
+                        // DEL will return the number of deleted keys, but we don't need them
+                        let _ = redis::cmd("DEL")
```
elpiel: just in case we have a "wrong" value in `Redis`, we are deleting it as a precaution.
Do you think this is necessary @Ivshti @samparsky ?

sentry/src/middleware/auth.rs
```diff
@@ -0,0 +1,95 @@
...
+                    .arg(serde_json::to_string(&session)?)
+                    .query_async(&mut redis.clone())
+                    .await?;
+
+                session
+            }
+        };
+        let ip = get_request_ip(&req);
+
+        let session = Session {
```
elpiel: The Application Session which includes the Request IP

sentry/src/middleware/auth.rs
```diff
@@ -0,0 +1,95 @@
...
+            era: adapter_session.era,
+            uid: adapter_session.uid.to_hex_non_prefix_string(),
+            ip,
+        };
+
+        req.extensions_mut().insert(session);
+    }
+
+    // @TODO: Check if we actually need this since we have the `adapter` for the check: `channelIfActive`
+    req.extensions_mut().insert(adapter.whoami().clone());
```
elpiel: Do we need this @samparsky @Ivshti ?

sentry/src/middleware/auth.rs
```diff
@@ -0,0 +1,95 @@
...
+                    Err(serde_error) => {
+                        // On a Deserialization error remove the key from Redis as a precaution
+                        // DEL will return the number of deleted keys, but we don't need them
+                        let _ = redis::cmd("DEL")
+                            .arg(token)
+                            .query_async::<_, u8>(&mut redis.clone())
+                            .await
+                            .map_err(|remove_error| {
+                                // @TODO: Consider logging this error instead.
+                                println!("{}", remove_error);
```
elpiel: If it is necessary, should we log this error @samparsky @Ivshti ?

sentry/src/middleware/auth.rs
```diff
@@ -0,0 +1,95 @@
...
+    req.extensions_mut().insert(adapter.whoami().clone());
+
+    Ok(req)
+}
+
+fn get_request_ip(req: &Request<Body>) -> Option<String> {
+    req.headers()
+        .get("true-client-ip")
+        .or_else(|| req.headers().get("x-forwarded-for"))
+        .and_then(|hv| hv.to_str().map(ToString::to_string).ok())
```
elpiel: FYI: `to_str()` returns a Result!

sentry/src/middleware/cors.rs
```diff
@@ -0,0 +1,164 @@
+use hyper::header::{ACCESS_CONTROL_ALLOW_HEADERS, ACCESS_CONTROL_REQUEST_HEADERS};
+use hyper::{Body, HeaderMap, Method, Request, Response, StatusCode};
+
+pub enum CorsResult {
```
elpiel: This is easier than using `Option<CorsResult>` since we don't need to unwrap the `Option` and add a dummy `Simple` response, we can directly `match` against `CorsResult`!

sentry/src/middleware/cors.rs
```diff
@@ -0,0 +1,164 @@
...
+        headers.insert(CONTENT_LENGTH, 0.into());
+        // if the request has `ACCESS_CONTROL_REQUEST_HEADERS` it is required to set `ACCESS_CONTROL_ALLOW_HEADERS`
+        if let Some(allow_headers) = req.headers().get(ACCESS_CONTROL_REQUEST_HEADERS) {
+            headers.insert(ACCESS_CONTROL_ALLOW_HEADERS, allow_headers.clone());
+        }
+
+        let mut response = Response::builder()
+            .status(StatusCode::NO_CONTENT)
+            .body(Body::empty())
+            .unwrap();
```
elpiel: this cannot fail atm.

sentry/src/middleware/auth.rs
samparsky: I don't think this is necessary Redis would overwrite values with the same key with the newest value

elpiel: It can be, because if it errors after the deserialization you won't get to the `SET`...

elpiel: This is resolved. If there is a wrong value in Redis, we will try to fetch the token again with the adapter and override the Redis value with it.

sentry/src/middleware/cors.rs
samparsky: I am not sure it's advisable to go against normal conventions. Having `Option<CorsResult>` makes it easy for anyone new to the codebase to understand

sentry/src/middleware/auth.rs
samparsky: Nope, I don't think so

elpiel: Removed.

sentry/src/lib.rs
```diff
@@ -89,18 +108,41 @@ where
...
 ) -> Response<Body> {
-    match (req.uri().path(), req.method()) {
+    let headers = match cors(&req) {
+        CorsResult::Simple(headers) => headers,
+        // if we have a Preflight, just return the response directly
+        CorsResult::Preflight(response) => return response,
+        CorsResult::None => Default::default(),
+    };
+
+    // otherwise problems with `.await` occurs about `Sync` being required for `Adapter`.
```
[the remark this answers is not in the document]
elpiel: Fixed.

sentry/src/lib.rs
```diff
@@ -89,18 +108,41 @@ where
...
+    let auth_connections = (adapter.clone(), redis.clone());
+    let req = match auth::for_request(req, &auth_connections.0, auth_connections.1).await {
+        Ok(req) => req,
+        Err(response_error) => return map_response_error(response_error),
```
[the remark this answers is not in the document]
elpiel: Yes, you are right! But this is error occurred if an Auth header was sent, it's not checking if it is required or not for the route.

sentry/src/middleware/auth.rs
```diff
@@ -0,0 +1,155 @@
...
+    let authorization = req.headers().get(AUTHORIZATION);
+
+    let prefix = "Bearer ";
+
+    let token = authorization
+        .and_then(|hv| {
+            hv.to_str()
+                .map(|token_str| {
+                    if token_str.starts_with(prefix) {
+                        Some(token_str.replacen(prefix, "", 1))
```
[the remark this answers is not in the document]
elpiel: Fixed.

sentry/src/middleware/cors.rs
elpiel: Removed.

sentry/src/lib.rs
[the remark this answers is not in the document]
samparsky: This is could be a security risk if the error trace contains some specific information that shouldn't be public. We should use a custom error message here

PR #186, opened 2019-11-08
Fix game theoretical failure

Ivshti:
Other than fixing the game theoretical failure described in https://github.com/AdExNetwork/adex-validator/issues/47, this PR also:

* fixes a small but important difference: health threshold is checked via GTE in JS, but only GT here (>= vs >)
* the tests could be made to check the exact outputs of `get_health`, like the JS tests

Other than that, it mirrors https://github.com/AdExNetwork/adex-validator/pull/227 and passes the JS integration tests so it should be good to go

### DIFF
--- a/validator_worker/src/follower.rs
+++ b/validator_worker/src/follower.rs
@@ -4,7 +4,7 @@ use primitives::adapter::Adapter;
 use primitives::validator::{ApproveState, MessageTypes, NewState, RejectState};
 use primitives::BalancesMap;
 
-use crate::core::follower_rules::{is_healthy, is_valid_transition};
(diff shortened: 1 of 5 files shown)


validator_worker/src/follower.rs
```diff
@@ -14,6 +14,7 @@ enum InvalidNewState {
     RootHash,
     Signature,
     Transition,
+    Health,
```
elpiel: Should we add a doc comment, which states this is for the Unsignable Heath InvalidNewState

Ivshti: we can rename it to UnsignableHealth if you want

Issue #189, opened 2019-11-12
Channel Middlewares


elpiel:
### [JS impl](https://github.com/AdExNetwork/adex-validator/blob/05-stable/middlewares/channel.js) reference

- [x] channelLoad (channel_load) #198 
- [x] channelIfExists (channel_load) #198 
- [x] channelIfFind (private) (channel_load) #198 
- [x] channelIfActive  #198

PR #191, opened 2019-11-13
Issue #189 Channel load

elpiel:
Part of #189 and prepares the code for the middleware `channelLoad`


primitives/src/channel.rs
```diff
@@ -10,11 +10,13 @@ use crate::big_num::BigNum;
 use crate::util::serde::ts_milliseconds_option;
 use crate::{AdUnit, EventSubmission, TargetingTag, ValidatorDesc, ValidatorId};
 
+pub type ChannelId = [u8; 32];
```
elpiel: type alias to make it more easy

sentry/src/db.rs
```diff
@@ -2,12 +2,20 @@ use redis::aio::MultiplexedConnection;
...
-    tokio_postgres::{Error as PostgresError, NoTls},
-    PostgresConnectionManager,
-};
+use bb8_postgres::tokio_postgres::NoTls;
+use bb8_postgres::PostgresConnectionManager;
 use lazy_static::lazy_static;
 
+pub type DbPool = Pool<PostgresConnectionManager<NoTls>>;
+
+pub mod field {
```
elpiel: In order not to add the whole `tokio_postgres` in primitives. We might add it as a feature, but don't think we should do it as it specifically for sentry and it's persistent mechanism and DB.

sentry/src/db/field/bignum.rs
```diff
@@ -0,0 +1,55 @@
...
+#[derive(Debug)]
+pub(crate) struct BigNumPg(BigNum);
+
+impl Into<BigNum> for BigNumPg {
+    fn into(self) -> BigNum {
+        self.0
+    }
+}
+
+impl<'a> FromSql<'a> for BigNumPg {
```
elpiel: And since we can't `impl Trait` for types defined in another crate with Trait being a trait from another crate, we need this wrapper.

sentry/src/middleware/channel.rs
```diff
@@ -0,0 +1,43 @@
...
+use primitives::{Channel, ChannelSpec};
+use std::error;
+
+pub async fn channel_load(
+    mut req: Request<Body>,
+    (pool, id): (DbPool, &ChannelId),
+) -> Result<Request<Body>, Box<dyn error::Error>> {
+    let channel_id = hex::encode(id);
+
+    let channel = pool
```
elpiel: We should probably move this to more general-purpose place, instead of leaving it here so we can reuse it.
I am opened for suggestions, but this is **not** part of this PR.

sentry/src/middleware/channel.rs
```diff
@@ -0,0 +1,43 @@
...
+            }
+        })
+        .await?;
+
+    req.extensions_mut().insert(channel);
+
+    Ok(req)
+}
+
+fn channel_map(row: &Row) -> Channel {
```
elpiel: Probably that should go out with the fetching of channel(s). See comment above.

sentry/src/db.rs
samparsky: this sounds okay

sentry/src/middleware/channel.rs
elpiel: It's just easier to show you the error:

```
error[E0117]: only traits defined in the current crate can be implemented for arbitrary types
  --> sentry/src/middleware/channel.rs:34:1
   |
34 | impl From<Row> for Channel {
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^ impl doesn't use types inside crate
   |
   = note: the impl does not reference only types defined in this crate
   = note: define and implement a trait or new type instead

error: aborting due to previous error
```

If we keep the `field` module inside `sentry`, this cannot be done, since you cannot impl traits from one crate for types in another.

elpiel: > Also, `get` can panic should we prefer `get_opt`?
Good point, there is also `try_get`, but they panic in the cases of:
`Panics if the index is out of bounds or if the value cannot be converted to the specified type.`
Which means Dev mistake, which means it should panic in that case. We should handle the error.

samparsky: The From<From> is for ChannelPg

sentry/src/db/field/channel_id.rs
```diff
@@ -0,0 +1,65 @@
...
+        match *ty {
+            Type::TEXT | Type::VARCHAR => true,
+            _ => false,
+        }
+    }
+}
+
+impl ToSql for ChannelIdPg {
+    fn to_sql(&self, ty: &Type, w: &mut BytesMut) -> Result<IsNull, Box<dyn Error + Sync + Send>> {
+        let string = hex::encode(&self.0);
```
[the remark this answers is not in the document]
elpiel: haven't tried that, I might try it, it was easier to do it this way. And I am not sure how well it will play as a `primary key`. I also think I was looking into that a long long time ago (when I actually implemented this), but don't remember what happened.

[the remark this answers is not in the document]
samparsky: Storing as  binary data is actually more performant when to string

primitives/src/postgres/field/channel_id.rs
```diff
@@ -0,0 +1,32 @@
+use std::error::Error;
+
+use crate::ChannelId;
+use hex::FromHex;
+use postgres_types::{FromSql, Type};
+
+#[derive(Debug)]
+pub struct ChannelIdPg(ChannelId);
+
+impl Into<ChannelId> for ChannelIdPg {
```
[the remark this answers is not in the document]
elpiel: Fixed.

PR #193, opened 2019-11-18
Issue #151 migrations

### LINKED ISSUE #151: Database setup and migrations
Database querying and etc:
- :negative_squared_cross_mark: `diesel` - excluded as an option as it's a synchronous ORM.
- :question: `bb8` - there is currently ongoing work to move to async/await and they are talking about (maybe) improving the API https://github.com/khuey/bb8/pull/37

Migration crates:
- https://github.com/jaemk/migrant_lib
- https://github.com/rust-db/barrel

elpiel:
Resolves #151


sentry/src/main.rs
```diff
@@ -98,6 +98,8 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
 
     let logger = logger();
     let redis = redis_connection().await?;
+    // setup migrations before setting up Postgres
+    setup_migrations().await;
```
elpiel: `bb8` doesn't  check if the credentials are correct, I haven't diged into this, but it usually times out on the first request, so it's good that we first setup the migrations which will check the credentials.

sentry/src/db.rs
```diff
@@ -1,18 +1,28 @@
...
+        env::var("POSTGRES_USER").unwrap_or_else(|_| String::from("postgres"));
+    static ref POSTGRES_PASSWORD: String =
+        env::var("POSTGRES_PASSWORD").unwrap_or_else(|_| String::from("postgres"));
+    static ref POSTGRES_HOST: String =
+        env::var("POSTGRES_HOST").unwrap_or_else(|_| String::from("localhost"));
+    static ref POSTGRES_PORT: u16 = env::var("POSTGRES_PORT")
+        .unwrap_or("5432".to_string())
+        .parse()
+        .unwrap();
+    static ref POSTGRES_DB: Option<String> = env::var("POSTGRES_DB").ok();
```
samparsky: Wouldn't it be much better to have POSTGRES_URL as we did before and parse it for `migrant`. Having to pass several environment variables could become unfriendly quickly.

elpiel: Well then we need to copy-paste this in that case:
https://github.com/sfackler/rust-postgres/blob/master/tokio-postgres/src/config.rs#L667

elpiel: I don't want to do this, let's keep it simple and for now have separate env. variables. We can always improve later. `Migrant` is not helping as well as this is a result of it + tokio_postgres doesn't have the properties of the config publicly.

sentry/migrations/20190806011140_initial_tables/up.sql
```diff
@@ -0,0 +1,39 @@
+CREATE TABLE channels
+(
+    channel_id     VARCHAR(66)              NOT NULL,
```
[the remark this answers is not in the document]
elpiel: It can be whatever we call it.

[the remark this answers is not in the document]
elpiel: Fixed.

PR #203, opened 2019-11-25
Issue #9 channel list

elpiel:
Part of #9


sentry/src/db/channel.rs
```diff
@@ -67,3 +82,93 @@ pub async fn insert_channel(
...
+                        }
+                    },
+                    Err(e) => Err((e, connection)),
+                }
+            }
+        })
+        .await
+}
+
+pub async fn list_channels_total_pages(
```
elpiel: Just realized that this is counting all records, which means that we need to calculate the TotalPages in the route handler since it should be calculated accordingly to the limit per page.

sentry/src/db/channel.rs
```diff
@@ -67,3 +82,93 @@ pub async fn insert_channel(
...
+        serde_json::Value::from_str(&format!(r#"[{{"id": "{}"}}]"#, validator_id))
+            .expect("Not a valid json")
+    });
+    let (where_clauses, params) =
+        channel_list_query_params(creator, validator.as_ref(), valid_until_ge);
+
+    pool.run(move |connection| {
+        async move {
+            let statement = format!(
+                "SELECT COUNT(id)::varchar FROM channels WHERE {}",
```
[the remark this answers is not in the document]
elpiel: There was an error that `int8` (which was returned by default from the `COUNT(id)`) is not a valid number or smth like that, which was very strange, because there were no invalid characters and it should have worked with it, but it didn't. That's why I used varchar and then either way we cast it from String to u64 in `TotalPages`

sentry/src/db/channel.rs
```diff
@@ -67,3 +82,93 @@ pub async fn insert_channel(
...
+            .expect("Not a valid json")
+    });
+    let (where_clauses, params) =
+        channel_list_query_params(creator, validator.as_ref(), valid_until_ge);
+
+    pool
+        .run(move |connection| {
+            async move {
+                // To understand why we use Order by, see Postgres Documentation: https://www.postgresql.org/docs/8.1/queries-limit.html
+                let statement = format!("SELECT id, creator, deposit_asset, deposit_amount, valid_until, spec FROM channels WHERE {} ORDER BY id DESC LIMIT {} OFFSET {}", where_clauses.join(" AND "), limit, skip);
```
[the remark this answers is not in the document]
elpiel: I guess we can. We need to probably add an index in that case.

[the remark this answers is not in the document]
elpiel: Order added!

sentry/src/db/channel.rs
elpiel: This is fixed now.

sentry/migrations/20190806011140_initial-tables/up.sql
```diff
@@ -3,37 +3,33 @@ CREATE TABLE channels
...
-    deposit_amount VARCHAR(255)             NOT NULL, -- @TODO change the deposit to BigNum compatible field
+    deposit_amount VARCHAR(255)             NOT NULL,
     valid_until    TIMESTAMP WITH TIME ZONE NOT NULL,
     spec           JSONB                    NOT NULL,
 
     PRIMARY KEY (id)
 );
 
-CREATE INDEX idx_valid_until ON channels (valid_until);
-CREATE INDEX idx_spec ON channels ((spec -> 'validator' ->> 'id'));
```
elpiel: We should see if this is even possible with value in array without using heavy indices like these:
https://www.postgresql.org/docs/9.4/datatype-json.html#JSON-CONTAINMENT

sentry/src/db/channel.rs
```diff
@@ -67,3 +86,112 @@ pub async fn insert_channel(
         })
         .await
 }
+
+#[derive(Debug, Deserialize)]
+pub struct ListChannels {
+    pub total_count: u64,
+    pub channels: Vec<Channel>,
+}
```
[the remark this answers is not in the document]
elpiel: Yep, I did remove it and also move those fn in separate module to be self-contained.

sentry/src/db/channel.rs
```diff
@@ -67,3 +69,122 @@ pub async fn insert_channel(
...
+
+        // Use a varchar or text, since otherwise `int8` fails deserialization
+        accepts!(VARCHAR, TEXT);
+    }
+
+    #[derive(Debug)]
+    pub struct ListChannels {
+        pub total_count: u64,
+        pub channels: Vec<Channel>,
+    }
```
samparsky: ListChannels is the same as `ChannelListResponse`

sentry/src/db/channel.rs
```diff
@@ -67,3 +69,122 @@ pub async fn insert_channel(
...
+            }
+        })
+        .await
+    }
+
+    fn channel_list_query_params<'a>(
+        creator: &'a Option<String>,
+        validator: Option<&'a serde_json::Value>,
+        valid_until_ge: &'a DateTime<Utc>,
+    ) -> (Vec<String>, Vec<&'a (dyn ToSql + Sync)>) {
```
elpiel: We can use HashMap later, this made it easier to finish up the code, but it's an improvement we can do. There might be complications with the parameters when we need them for the prepared statements.

sentry/src/db/channel.rs
elpiel: One returns the `total_count`, while the other returns the `total_pages`, but we can move the calculation of the `total_pages` here and will change it to `ChannelListReponse`

PR #209, opened 2019-11-28
Analytics route

samparsky:
Part of #9


sentry/src/routes/analytics.rs
```diff
@@ -0,0 +1,236 @@
...
+    pub timeframe: String,
+}
+
+impl AnalyticsQuery {
+    pub fn is_valid(&self) -> Result<(), ResponseError> {
+        let valid_event_types = ["IMPRESSION"];
+        let valid_metric = ["eventPayouts", "eventCounts"];
+        let valid_timeframe = ["year", "month", "week", "day", "hour"];
+
+        if !valid_event_types.iter().any(|e| *e == &self.event_type[..]) {
```
elpiel: you can just compare `String`s, why so complicated?
Same applies for the checks bellow.

sentry/src/routes/analytics.rs
```diff
@@ -0,0 +1,236 @@
...
+pub async fn process_analytics<A: Adapter>(
+    req: Request<Body>,
+    app: &Application<A>,
+    is_advertiser: bool,
+    skip_publisher_filter: bool,
+) -> Result<String, ResponseError> {
+    let query = serde_urlencoded::from_str::<AnalyticsQuery>(&req.uri().query().unwrap_or(""))?;
+    query.is_valid()?;
+
+    let applied_limit = cmp::min(query.limit, 200);
```
elpiel: Tip: you can just call `query.limit.min()`

sentry/src/routes/analytics.rs
```diff
@@ -0,0 +1,236 @@
...
+    let (interval, period) = get_time_frame(&query.timeframe);
+    let time_limit = Utc::now().timestamp() - period;
+    let sess = req.extensions().get::<Session>();
+
+    let mut where_clauses = vec![format!("created > to_timestamp({})", time_limit)];
```
elpiel: I would like to keep all querying stuff separated in the db module.

sentry/src/db/channel.rs
```diff
@@ -70,6 +70,27 @@ pub async fn insert_channel(
...
 }
 
+pub async fn get_channel_by_creator(
+    pool: &DbPool,
+    id: &ChannelId,
+) -> Result<Vec<Channel>, RunError<bb8_postgres::tokio_postgres::Error>> {
+    pool.run(move |connection| {
+        async move {
+            match connection
+                .prepare("SELECT id FROM channels WHERE creator = $1")
```
elpiel: There is no limit? We don't even use `ChannelId`? Should this get single or multiple channels? it says `channel` in the function so I expect it will be single.

samparsky: Removed it as it's not needed

sentry/src/routes/analytics.rs
samparsky: It wouldn't work because it would be comparing &str with String. I made the change to `.contains` though

sentry/src/routes/analytics.rs
samparsky: Sounds okay

sentry/src/routes/analytics.rs
```diff
@@ -0,0 +1,97 @@
...
+) -> Result<Response<Body>, ResponseError> {
+    let request_uri = req.uri().to_string();
+    let redis = app.redis.clone();
+
+    match redis::cmd("EXISTS")
+        .arg(&request_uri)
+        .query_async::<_, String>(&mut redis.clone())
+        .await
+    {
+        Ok(response) => Ok(success_response(response)),
```
elpiel: `Exists` returns a string representation of either `0` or `1` see https://redis.io/commands/exists#return-value

Which means that you probably should probably check for that otherwise the next matching arm doesn't make any sense as it will only work if there is an Error that occurred trying to reach `redis` or executing the command.

Edit: Apparently this should be a `GET`, so keep in mind the return values of `GET`, see https://redis.io/commands/get#return-value

sentry/src/db/analytics.rs
```diff
@@ -0,0 +1,95 @@
...
+use chrono::Utc;
+use primitives::analytics::{AnalyticsQuery, AnalyticsResponse};
+
+pub async fn get_analytics(
+    query: AnalyticsQuery,
+    channels: Option<String>,
+    sess: Option<&Session>,
+    pool: &DbPool,
+    is_advertiser: bool,
+    skip_publisher_filter: bool,
```
elpiel: Better to fix those 2 parameters as they, for what I can see, are bounded to the `Session` and we do an `unwrap` on the session, which might be wrongly passed. I feel that this approach resembles more JS style than what we can do with Rust.

We can use a Struct or Enum to represent those filters and include the Session and handle the cases inside.

Also `skip_publisher_filter`, as general rule of thumb for me, should be renamed to something more clean for the API like `filter_publisher`.

primitives/src/analytics.rs
```diff
@@ -0,0 +1,78 @@
...
+    #[serde(default = "default_event_type")]
+    pub event_type: String,
+    #[serde(default = "default_metric")]
+    pub metric: String,
+    #[serde(default = "default_timeframe")]
+    pub timeframe: String,
+}
+
+impl AnalyticsQuery {
+    pub fn is_valid(&self) -> Result<(), DomainError> {
```
elpiel: Instead of using strings, it will be better to use Enums and leave this kind of validation to Serde.
Probably also we don't want a big limit as it might cause the DB to hang, we should think of a good max limit to put here.

sentry/src/db/analytics.rs
```diff
@@ -0,0 +1,95 @@
+use crate::db::DbPool;
+use crate::Session;
+use bb8::RunError;
+use chrono::Utc;
+use primitives::analytics::{AnalyticsQuery, AnalyticsResponse};
+
+pub async fn get_analytics(
+    query: AnalyticsQuery,
+    channels: Option<String>,
```
elpiel: Can't we just deserialize this? It's very hard to reason about this `String` value. Also apparently it's a single `channel`

primitives/src/analytics.rs
samparsky: I don't think using enum offer any benefits at all. 
We would have to derive display so to call `to_string` on the enum, then the errors `serde` would return wouldn't make a lot of sense so we have to implement custom errors. Seem like an unneeded implementation detail

samparsky: Also, 100-200 (max) is a good limit. Postgres shouldn't hang on that if we have good indexes

sentry/src/db/analytics.rs
samparsky: There is no point to deserialize, it either an appropriate value that returns some data or nothing

sentry/src/db/analytics.rs
samparsky: `Better to fix those 2 parameters as they` ... I don't get this

sentry/src/db/analytics.rs
elpiel: Ok, that sounds right. Can we at least rename it to `channel` as it's only querying a single channel

sentry/src/db/analytics.rs
elpiel: in a nutshell, if I pass `None` & `is_advertiser: false`, this will blow up. We need a better API for this function.

primitives/src/analytics.rs
elpiel: yes, but since this comes from deserialization of the query passed from the URL, you could pass 1 million and then it will hang, i.e. we need validation of the limit.

samparsky: 1 million wouldn't work because we have a max limit of 200 https://github.com/AdExNetwork/adex-validator-stack-rust/pull/209/files#diff-3870d7b9b77c57e1688961c8a3f97b50R15

elpiel: Can you move this to the Query and add a check in `is_valid` instead as this hardcoded value on the DB query side can lead to confusion as it's hidden and doesn't give any feedback that it actually limited the value to `200`.

sentry/src/db/analytics.rs
```diff
@@ -0,0 +1,117 @@
+use crate::db::DbPool;
+use crate::Session;
+use bb8::RunError;
+use chrono::Utc;
+use primitives::analytics::{AnalyticsQuery, AnalyticsResponse, ANALYTICS_QUERY_LIMIT};
+
+pub enum AnalyticsType {
```
elpiel: That looks much better. Thank you for the effort!

sentry/src/routes/analytics.rs
```diff
@@ -0,0 +1,116 @@
...
+    logger: &Logger,
+) {
+    if let Err(err) = redis::cmd("SETEX")
+        .arg(&key)
+        .arg(timeframe)
+        .arg(value)
+        .query_async::<_, ()>(&mut redis.clone())
+        .await
+    {
+        error!(&logger, "server error: {}", err);
```
elpiel: We should improve the logging later on with `module` key-value pair

PR #229, opened 2020-01-03
Issue #9 GET Channel Events aggregates

elpiel:
part of #9

### DIFF
--- a/sentry/src/routes/event_aggregate.rs
+++ b/sentry/src/routes/event_aggregate.rs
@@ -1,24 +1,47 @@
-use crate::{Application, ResponseError, Session};
+use chrono::{serde::ts_milliseconds_option, DateTime, Utc};
 use hyper::{Body, Request, Response};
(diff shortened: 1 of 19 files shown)


sentry/src/routes/event_aggregate.rs
```diff
@@ -0,0 +1,47 @@
...
+}
+
+pub async fn list_channel_event_aggregates<A: Adapter>(
+    req: Request<Body>,
+    app: &Application<A>,
+) -> Result<Response<Body>, ResponseError> {
+    let channel = req
+        .extensions()
+        .get::<Channel>()
+        .expect("Request should have Channel");
```
samparsky: Should it crash here or return an error?

elpiel: Crash. As we don't have a direct way to pass a loaded channel from the `load_channel` middleware :(

PR #230, opened 2020-01-03
Update validator_worker dependencies

elpiel:
Updating to stable releases :tada:


validator_worker/src/sentry_interface.rs
```diff
@@ -83,7 +82,7 @@ impl<T: Adapter + 'static> SentryApi<T> {
         }))
         .await
         {
-            handle_http_error(e);
+            println!("Propagation error: {}", e);
```
samparsky: let's log the error here

elpiel: Since we don't have the logging setup in `validator_worker` and it's not as straight forward as I though and nothing to do with the dependencies update I would like to do this in separate PR.

PR #238, opened 2020-01-13
Issue #233 validator worker docker image

### LINKED ISSUE #233: Docker image
Docker Image:
- ~~`alpine` linux~~ (not possible atm)
- The `validator_worker` binary

Should:
- Be compatible with the JS impl & environment variables

`alpine` is not possible at the moment since:
```
Caused by:
  cannot produce proc-macro for `const-random-macro v0.1.6` as the target `x86_64-unknown-linux-musl` does not support these crate types
```

elpiel:
Fixes #233

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -32,20 +32,14 @@ ENV SENTRY_URL=
 # default: `infinite`
 ENV SINGLE_TICK=
 
-RUN echo ${KEYSTORE_FILE:+-k $KEYSTORE_FILE} ${SENTRY_URL:+-u $SENTRY_URL}
-
 WORKDIR /usr/local/bin
 
 RUN apt update && apt-get install -y libssl-dev
 
 COPY --from=builder /usr/local/bin/validator_worker .
 
-ENTRYPOINT [ \
-    "sh", "-c", \
-    "validator_worker", \
-    "-a", "${ADAPTER:-ethereum}", \
-    "${DUMMY_IDENTITY:+-i $DUMMY_IDENTITY}", \
-    "${KEYSTORE_FILE:+-k $KEYSTORE_FILE}", \
-    "${SINGLE_TICK:+-t}", \
-    "${SENTRY_URL:+-u $SENTRY_URL}", "${CONFIG}" \
-]
+ENTRYPOINT validator_worker -a ${ADAPTER:-ethereum} \
+            ${KEYSTORE_FILE:+-k $KEYSTORE_FILE} \
+            ${DUMMY_IDENTITY:+-i $DUMMY_IDENTITY} \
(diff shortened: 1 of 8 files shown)


adapter/Cargo.toml
```diff
@@ -19,8 +19,8 @@ eth_checksum = "0.1.1"
 ethabi = { git = "https://github.com/samparsky/ethabi", branch = "graph-patches" }
 tiny-keccak = "1.5"
 parity-crypto = { version = "0.4.2", features = ["publickey"] }
-ethstore = { version = "0.2.1", git = "https://github.com/paritytech/parity-ethereum"}
-ethkey = { version = "0.4.0", git = "https://github.com/paritytech/parity-ethereum"}
+ethstore = { version = "0.2.1", git = "https://github.com/elpiel/parity-ethereum"}
+ethkey = { version = "0.4.0", git = "https://github.com/elpiel/parity-ethereum"}
```
samparsky: why the change here?

Dockerfile
```diff
@@ -0,0 +1,45 @@
+FROM rust:1.40 as builder
+
```
samparsky: Nice work! using multi-stage builds

Dockerfile
```diff
@@ -0,0 +1,45 @@
...
+COPY . .
+
+RUN cargo install --path validator_worker --all-features
+
+WORKDIR /usr/local/bin
+
+RUN cp $CARGO_HOME/bin/validator_worker .
+
+FROM ubuntu:18.04
```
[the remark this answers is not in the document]
Ivshti: yeah, alpine uses musl so it can't run the rust binaries
@elpiel correct me if I'm wrong

adapter/Cargo.toml
Ivshti: does it sync faster?

Dockerfile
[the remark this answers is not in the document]
elpiel: Yes, that's the problem (`openssl` crate doesn't build on musc)

adapter/Cargo.toml
elpiel: faster - yes
fast build - not so much it still takes long time, but it's better.
I tried looking into caching dependencies, but not much luck on that.
@samparsky sent me a link, so I will check it if it's possible to use cache for the **build**

Dockerfile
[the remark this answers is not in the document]
elpiel: Sorry, I lied... it's not `openssl`, it was another crate:
```
Caused by:
  cannot produce proc-macro for `const-random-macro v0.1.6` as the target `x86_64-unknown-linux-musl` does not support these crate types
```

PR #239, opened 2020-01-16
Add event routes

samparsky:
/:id/events route POST
/:id/events-aggregates GET 

`sample response /:id/events-aggregates GET  `

```js
{
channel: {},
"events": [
        {
            "channelId": "0x061d5e2a67d0a9a10f1c732bca12a676d83f79663a396f7d87b3e30b9b411088",
            "created": "2020-01-24T15:13:36.218136Z",
            "events": {
                "IMPRESSION": {
                    "eventCounts": {
                        "0xB7d3F81E857692d13e9D63b232A90F4A1793189E": "1",
                        "0x759959cD77962981921009da6671b1cbEE5203C9": "2"
                    },
                    "eventPayouts": {
                        "0x759959cD77962981921009da6671b1cbEE5203C9": "2",
                        "0xB7d3F81E857692d13e9D63b232A90F4A1793189E": "1"
                    }
                }
            }
        }
    ]
}
```

#242


sentry/src/middleware/auth.rs
```diff
@@ -75,6 +75,7 @@ fn get_request_ip(req: &Request<Body>) -> Option<String> {
         .get("true-client-ip")
         .or_else(|| req.headers().get("x-forwarded-for"))
         .and_then(|hv| hv.to_str().map(ToString::to_string).ok())
+        .map(|token| token.split(',').collect::<Vec<&str>>()[0].to_string())
```
elpiel: `split()` always returns an element (even if it's empty one), so you can `token.split(',').nth(1).expect("should always be here")` rather than collecting in a `Vec` just to take the first one out.

sentry/src/access.rs
```diff
@@ -2,29 +2,43 @@ use chrono::Utc;
...
 use primitives::Channel;
 use std::cmp::PartialEq;
-
-use crate::Session;
+use std::error::Error;
+use std::fmt;
 
 #[derive(Debug, PartialEq, Eq)]
-pub enum Error {
+pub enum AccessError {
```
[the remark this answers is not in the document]
samparsky: I changed the name because of `use std::error::Error;`

sentry/src/db/event_aggregate.rs
```diff
@@ -52,3 +52,73 @@ pub async fn list_event_aggregates(
...
+    channel_id: &ChannelId,
+    event: &EventAggregate,
+) -> Result<bool, RunError<bb8_postgres::tokio_postgres::Error>> {
+    let mut values = Vec::new();
+    let mut index = 0;
+    let id = channel_id.to_string();
+
+    let mut data: Vec<String> = Vec::new();
+
+    for (event_type, aggr) in &event.events {
```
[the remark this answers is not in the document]
samparsky: Yeah, this is right I have changed the query. Although the probability of us hitting the query limit is low

sentry/src/db/event_aggregate.rs
```diff
@@ -52,3 +52,73 @@ pub async fn list_event_aggregates(
...
+        if let Some(event_counts) = &aggr.event_counts {
```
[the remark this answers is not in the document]
samparsky: I am not sure what you mean, I am using  the `aggr`  var

sentry/src/event_aggregator.rs
```diff
@@ -0,0 +1,121 @@
...
+    }
+}
+
+async fn store(
+    db: &DbPool,
+    channel_id: &ChannelId,
+    aggr: Arc<RwLock<HashMap<String, EventAggregate>>>,
+) {
+    let mut recorder = aggr.write().await;
+    let ev_aggr: Option<&EventAggregate> = recorder.get(&channel_id.to_string());
```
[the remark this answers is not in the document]
samparsky: Because `ChannelId` doesn't implement `Hash`

sentry/src/routes/channel.rs
```diff
@@ -102,6 +103,35 @@ pub async fn last_approved<A: Adapter>(
...
+        .get::<Channel>()
+        .expect("Request should have Channel")
+        .to_owned();
+
+    let into_body = req.into_body();
+    let body = hyper::body::to_bytes(into_body).await?;
+    let events = serde_json::from_slice::<Vec<Event>>(&body)?;
+
+    app.event_aggregator
+        .record(app, channel, session, &events.as_slice())
```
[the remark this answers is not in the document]
samparsky: `req.into_body()` consumes the `req` var (i.e. owns it and all references to the data underlying become invalid). The rust compiler complains if there are existing references to the consumed `req` var, that's the reason behind transferring the ownership

sentry/src/access.rs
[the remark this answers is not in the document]
elpiel: What about it?

sentry/src/db/event_aggregate.rs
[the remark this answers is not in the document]
elpiel: Instead of `for` loop and then `if let` just filter the iterator and use the for loop on the results:
`for (...) in event.events.filter(Some) {`

sentry/src/event_aggregator.rs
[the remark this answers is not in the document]
elpiel: We can just derive it, it's not big deal.

sentry/src/db/event_aggregate.rs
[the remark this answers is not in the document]
samparsky: It's not a filter it's to prevent having to write `unwrap()` or `expect`

sentry/src/routes/channel.rs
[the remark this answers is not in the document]
elpiel: https://docs.rs/hyper/0.13.1/hyper/struct.Request.html#method.body
Although I am not sure if the `to_bytes` will work with a reference.

sentry/src/access.rs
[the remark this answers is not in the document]
samparsky: name collision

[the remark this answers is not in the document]
elpiel: All the crates do it and is acceptable. It doesn't add value to the PR.
https://docs.rs/reqwest/0.10.0/reqwest/struct.Error.html

sentry/src/db/event_aggregate.rs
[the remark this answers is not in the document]
elpiel: fair enough

sentry/src/access.rs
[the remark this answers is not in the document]
Ivshti: agreed

sentry/migrations/20190806011140_initial-tables/seed/up.sql
```diff
@@ -0,0 +1,11 @@
+INSERT INTO 
+channels 
+VALUES 
+(
+    '0x061d5e2a67d0a9a10f1c732bca12a676d83f79663a396f7d87b3e30b9b411088', 
+    '0x033ed90e0fec3f3ea1c9b005c724d704501e0196', 
+    '0x89d24A6b4CcB1B6fAA2625fE562bDD9a23260359', 
+    '1000',
```
Ivshti: is it possible to use the native bigint here?

sentry/migrations/20190806011140_initial-tables/up.sql
```diff
@@ -28,8 +28,19 @@ CREATE INDEX idx_validator_messages_msg_state_root ON validator_messages ((msg -
 CREATE TABLE event_aggregates
 (
     channel_id VARCHAR(66)              NOT NULL REFERENCES channels (id) ON DELETE RESTRICT,
-    created    TIMESTAMP WITH TIME ZONE NOT NULL,
-    events     JSONB                    NOT NULL
+    created    TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT NOW(),
+    event_type VARCHAR(255)             NOT NULL,
+    earner     VARCHAR(255),
+    event_counts VARCHAR                   NOT NULL,
```
Ivshti: can't we use the native bigint here? also, `count` and `payout` is sufficient in terms of naming

sentry/src/db/event_aggregate.rs
```diff
@@ -1,38 +1,72 @@
...
         .run(move |connection| {
             async move {
                 let where_clause = if !where_clauses.is_empty() {
-                    format!("WHERE {}", where_clauses.join(" AND "))
+                    where_clauses.join(" AND ").to_string()
                 } else {
                     "".to_string()
                 };
-                let statement = format!("SELECT channel_id, created, events FROM event_aggregates {} ORDER BY created DESC LIMIT {}", where_clause, limit);
+                let statement = format!(
```
Ivshti: I haven't reviewed this query in detail, but it looks very elegant :+1:

sentry/src/db/event_aggregate.rs
```diff
@@ -52,3 +86,88 @@ pub async fn list_event_aggregates(
...
+                    event_payout: event_payout.to_string(),
+                    created,
+                }]);
+
+                // total sum
+                total_event_counts = event_count.add(&total_event_counts);
+                total_event_payouts = total_event_payouts.add(event_payout);
+            }
+
+            data.extend(vec![EventData {
```
Ivshti: can't you just `push`?

sentry/src/db/event_aggregate.rs
```diff
@@ -1,38 +1,72 @@
...
+                let statement = format!(
+                    "
+                        WITH aggregates AS (
+                            SELECT 
+                                channel_id,
+                                created,
+                                event_type,
+                                jsonb_build_object(
+                                    'eventCounts',
+                                    jsonb_object_agg(
```
Ivshti: how does this `agg` actually sum the values in counts/payouts?

sentry/src/event_aggregator.rs
```diff
@@ -0,0 +1,127 @@
...
+                // the channel events to database
+                if aggr_throttle > 0 {
+                    tokio::spawn(async move {
+                        loop {
+                            // break loop if the
+                            // channel withdraw period has started
+                            // since no event is allowed once a channel
+                            // is in withdraw period
+
+                            if Utc::now() > withdraw_period_start {
```
Ivshti: this check should not be here because when throttle is set to 0, we will ignore it

it should be in the beginning of `record`

sentry/src/event_aggregator.rs
```diff
@@ -0,0 +1,127 @@
...
+
+pub fn new_aggr(channel_id: &ChannelId) -> EventAggregate {
+    EventAggregate {
+        channel_id: channel_id.to_owned(),
+        created: Utc::now(),
+        events: HashMap::new(),
+    }
+}
+
+async fn store(db: &DbPool, channel_id: &ChannelId, logger: &Logger, aggr: Aggregate) {
```
Ivshti: The naming (or something) is quite confusing here

in `record`, you get an `aggr` FROM the `recorder`, but here you somehow turn the aggr INTo a `recorder`? how does this work?

sentry/src/routes/channel.rs
```diff
@@ -102,6 +104,39 @@ pub async fn last_approved<A: Adapter>(
...
+
+    let into_body = req.into_body();
+    let body = hyper::body::to_bytes(into_body).await?;
+    let request_body = serde_json::from_slice::<HashMap<String, Vec<Event>>>(&body)?;
+    let events = request_body
+        .get("events")
+        .ok_or_else(|| ResponseError::BadRequest("invalid request".to_string()))?;
+
+    app.event_aggregator
+        .record(app, &channel, &session, &events)
```
Ivshti: In the JS implementation, to achieve max throughput we get the channel in the `recorder` itself. I suggest you use the same approach. There's no point to get the channel on every request here, when it can be retrieved when constructing the recorder, since it's immutable (even when we add dynamic data, it won't be all that important to update it with every request)

sentry/src/db/event_aggregate.rs
samparsky: It doesn't `SUM` the values, what it does it return aggr JSON similar to the implementation in JS. We don't `SUM` the aggr values in JS either.  At first I tried `SUM` but realized we can't `SUM` because we use the `aggr.created` time and `SUM` query won't return that because it would be across multiple times.

sentry/src/db/event_aggregate.rs
samparsky: Yeah true

sentry/src/db/event_aggregate.rs
```diff
@@ -1,38 +1,72 @@
...
+        where_clauses.push(format!("earner = '{}'", from.to_string()));
+        where_clauses.push("event_type = 'IMPRESSION'".to_string());
+    } else {
+        where_clauses.push("earner is NOT NULL".to_string());
     }
+
     if let Some(after) = after {
         params.push(after);
-        where_clauses.push(format!("created > {}", params.len()));
+        where_clauses.push(format!("created > ${}", params.len()));
```
elpiel: why don't we use the `params` for all query parameters but _only_ for `created > ${}`?

sentry/src/db/event_aggregate.rs
```diff
@@ -52,3 +86,88 @@ pub async fn list_event_aggregates(
...
+
+    let mut data: Vec<EventData> = Vec::new();
+
+    for (event_type, aggr) in &event.events {
+        if let Some(event_counts) = &aggr.event_counts {
+            let mut total_event_counts: BigNum = 0.into();
+            let mut total_event_payouts: BigNum = 0.into();
+            for (earner, event_count) in event_counts {
+                let event_payout = aggr.event_payouts[earner].clone();
+                data.extend(vec![EventData {
```
elpiel: `push()`?

sentry/src/db/event_aggregate.rs
```diff
@@ -52,3 +86,88 @@ pub async fn list_event_aggregates(
 
     Ok(event_aggregates)
 }
+
+#[derive(Debug)]
+struct EventData {
+    id: String,
+    event_type: String,
+    earner: Option<String>,
```
elpiel: We have `ToSql` for `ValidatorId` why not use it for this struct as well?

sentry/src/event_aggregator.rs
```diff
@@ -0,0 +1,127 @@
...
+    let mut recorder = aggr.write().await;
+    let ev_aggr: Option<&EventAggregate> = recorder.get(channel_id);
+    if let Some(data) = ev_aggr {
+        if let Err(e) = insert_event_aggregate(&db, &channel_id, data).await {
+            error!(&logger, "{}", e; "eventaggregator" => "store");
```
elpiel: `"module" => "event_aggregator", "in" => "store"`

sentry/src/event_aggregator.rs
samparsky: This check is meant to help kill the long-running asynchronous task, the check is already implemented in `check_access`

sentry/src/db/event_aggregate.rs
samparsky: Because `Postgres` doesn't recognize`ValidatorId` or `ChannelId`

sentry/src/event_aggregator.rs
samparsky: what's the difference?

elpiel: consistency?

samparsky: I prefer the short form

sentry/migrations/20190806011140_initial-tables/up.sql
samparsky: We can't use BIGINT because it can only represent 64 bits integers while we need 256 bits integer representation

sentry/migrations/20190806011140_initial-tables/seed/up.sql
samparsky: We can't use BIGINT because it can only represent 64 bits integers while we need 256 bits integer representation

sentry/src/event_aggregator.rs
elpiel: Then make an issue to fix _all_ the other places to make consistent + at least use `snake_case` so that we don't scratch our heads all the time

samparsky: https://github.com/AdExNetwork/adex-validator-stack-rust/blob/dev/sentry/src/main.rs#L148

elpiel: As I said - we should make them consistent. Also for the main, well it's not actually a module, it's the binary entry point. I am welcoming changes. Although I think a bit more verbosity is good for the logging if in the future we start logging other stuff (which are not modules, etc.) to keep us away from confusion

PR #240, opened 2020-01-16
Issue #235 checksummed validator

### LINKED ISSUE #235: Checksummed ValidatorId serialization
- Serialization of ChannelSpec should be checksummed
- Balance tree keys should be lowercased

elpiel:
Resolves #235 

- Added ToETHChecksum trait
- `BalancesMap` now has `ValidatorId` as key
- Event payouts and etc. also have `ValidatorId` as key

### DIFF
--- a/primitives/src/balances_map.rs
+++ b/primitives/src/balances_map.rs
@@ -1,19 +1,29 @@
 use std::collections::BTreeMap;
 
 use crate::BigNum;
+use serde_hex::{SerHexSeq, StrictPfx};
 use std::collections::btree_map::{Entry, Iter, Values};
 
-use serde::{ser::SerializeMap, Deserialize, Serialize, Serializer};
+use serde::{Deserialize, Serialize};
 use std::iter::FromIterator;
+use std::ops::Deref;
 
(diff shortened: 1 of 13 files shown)


primitives/src/balances_map.rs
```diff
@@ -1,5 +1,93 @@
...
+use std::collections::btree_map::{Entry, Iter, Values};
 
-pub type BalancesMap = BTreeMap<String, BigNum>;
+use serde::{ser::SerializeMap, Deserialize, Serialize, Serializer};
+use std::iter::FromIterator;
+use std::ops::Index;
+
+#[derive(Clone, Debug, Default, Deserialize, PartialEq, Eq)]
+#[serde(transparent)]
+pub struct BalancesMap(BTreeMap<ValidatorId, BigNum>);
```
elpiel: This is the simplest way of not implementing all the deserialization and serialization, as there is no way atm to specify you want the key of the `BTreeMap` (which already has `Serializer` implemented) to be lowercase or not.
We cannot override the `Serializer` of BTreeMap so we need a newtype for that!

PR #244, opened 2020-01-21
Issue #243 contrafactual accounts

### LINKED ISSUE #243: Contrafactual accounts
Regarding [this PR in the JS impl](https://github.com/AdExNetwork/adex-validator/pull/250) we need to make the appropriate changes by removing the Identity contract calls and use the relay instead.

elpiel:
Resolves #243 

Changes applied from: https://github.com/AdExNetwork/adex-validator/pull/250

### DIFF
--- a/adapter/src/ethereum.rs
+++ b/adapter/src/ethereum.rs
@@ -5,6 +5,7 @@ use ethkey::Password;
 use ethstore::SafeAccount;
 use futures::compat::Future01CompatExt;
 use futures::future::{BoxFuture, FutureExt};
+use futures::TryFutureExt;
 use lazy_static::lazy_static;
 use parity_crypto::publickey::{
     public_to_address, recover, verify_address, Address, Message, Signature,
@@ -15,6 +16,7 @@ use primitives::{
     config::Config,
     Channel, ToETHChecksum, ValidatorId,
 };
+use reqwest::Client;
 use serde::{Deserialize, Serialize};
 use serde_hex::{SerHexOpt, StrictPfx};
 use serde_json::Value;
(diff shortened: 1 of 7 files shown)


adapter/src/ethereum.rs
```diff
@@ -276,6 +266,49 @@ impl Adapter for EthereumAdapter {
...
+            .get(&relay_url)
+            .send()
+            .and_then(|res: Response| res.json())
+            .await
+            .map_err(|_| map_error("Fetching privileges failed"))?;
+
+        let privilege = identities_owned
+            .get(identity)
+            .copied()
+            .unwrap_or_else(|| 0_u8);
```
samparsky: It shouldn't return `0_u8` if it fails to unwrap it, it means the identity isn't allowed. it should return an error `insufficient privilege`

adapter/src/ethereum.rs
```diff
@@ -276,6 +266,49 @@ impl Adapter for EthereumAdapter {
...
+    pub fn new(relayer_url: &str) -> Result<Self, reqwest::Error> {
+        let client = Client::builder().build()?;
+
+        Ok(Self {
+            relayer_url: relayer_url.to_string(),
+            client,
+        })
+    }
+
+    pub async fn get_privilege(&self, identity: &[u8; 20]) -> Result<u8, AdapterError> {
```
samparsky: we should use `ValidatorId` for identity

adapter/src/ethereum.rs
```diff
@@ -276,6 +266,49 @@ impl Adapter for EthereumAdapter {
...
+        use reqwest::Response;
+        use std::collections::HashMap;
+
+        let relay_url = format!(
+            "{}/identity/by-owner/{}",
+            self.relayer_url,
+            hex::encode(identity)
```
samparsky: identity is `0x` prefixed

adapter/src/ethereum.rs
elpiel: `0` is no privileges, and if you've seen the code that uses this function, you will see that we check if it's `> 0`

adapter/src/ethereum.rs
elpiel: Sounds good!

adapter/src/ethereum.rs
elpiel: Yeah I am still testing it, couldn't do it in Thursday.

adapter/src/ethereum.rs
elpiel: We could make an enum, but not sure if you'll like that

samparsky: We don't need the privilege levels anymore, we can just return the error if we don't find it

adapter/src/ethereum.rs
elpiel: This is fixed! Also I had to pass the `VerifiedPayload.from` and not the identity here.

Issue #245, opened 2020-01-22
Avoid time-sensitive tests


elpiel:
Occasionally a test will fail because of timing which we currently don't control. In particular we're talking about [sentry/src/access.rs](https://github.com/AdExNetwork/adex-validator-stack-rust/blob/49b438d63f/sentry/src/access.rs#L153) tests.

elpiel:
Change of plan. We will just change the `RateLimit` to 1 millisecond instead 20k and that fixes the issue

PR #246, opened 2020-01-28
Validator messages

samparsky:
Adds /:id/validator-messages POST route


elpiel:
Build failed because of #245

PR #251, opened 2020-02-13
Issue #249 divisibility arbitrary percision

### LINKED ISSUE #249: payment spec: divisibility / arbitrary precision
We can implement a property of the payment spec that allows to divide a token to any arbitrary number of decimals, and use that unit internally

then, when the state tree is signed, it would be divided to get the real token unit (e.g. if we were dividing to `10**30`, and the real unit is `10**18`, we'd mul by `10**-12` and round down)

useful for AdExNetwork/adex-validator#14 

this can help going both ways - if the token multiplier is, say - 18, and we want to pay out `3*10**12`, there's no point to increment aggregates by `3*10**12`; we can just increment by 3 and set `10**12` to be the multiplier

## Precision

- **USD** - We can take a precision of `1 USD` to be `1 * 10**6` (i.e. 6 decimal places)
- **DAI** has a precision of `10**18`
- **USDT** (Tether) has a precision of `10**8`

This means that we can use a base of `10**12` for internal database storage, four couple of reasons:
1) CPM has _3_ decimals places precision, but that's for `1 000`, this means we need _3_ decimals places of precision more for a total of **6**
2) Since **DAI** has `10**18` and is the most precise of them, we can use it as a base, meaning we get the base

#### `10**12 = 10**18 - 10**6` (6 decimal points less precise than **DAI**)

This should only be applied to **Sentry** and it's internal representation in the Database of the payouts.

## Benefits

- easy support of multiple tokens cause we restrict the maximum precision 
- easier aggregation
- easier analytics

## Multiple tokes support

Combined with https://github.com/AdExNetwork/adex-validator/issues/324 we can have the whitelisted tokens

Sentry changes:
- [ ] Check on creation of `Channel` if the `depositAsset` is on the Whitelisted tokens
- [ ] Define a list of available **tokens** and their **precision** as well as their **chainId**
- [ ] Impl `ChannelSpec` `depositChainId` (optional) - if it's not specified, try to infer the chain ID from the `depositAsset` (token address)

## Related
- https://github.com/AdExNetwork/adex-validator/issues/324

elpiel:
Resolves #249


primitives/src/big_num.rs
```diff
@@ -7,8 +7,62 @@ use std::str::FromStr;
...
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct PrecisionU64(u64);
+
+impl Into<BigNum> for PrecisionU64 {
+    fn into(self) -> BigNum {
+        BigNum(GLOBAL_MULTIPLIER * self)
+    }
+}
+
+impl From<BigNum> for PrecisionU64 {
```
elpiel: Should we use `From` with default of `0` or should we impl `TryFrom` and fail if this conversion cannot be done?

primitives/src/big_num.rs
```diff
@@ -7,8 +7,62 @@ use std::str::FromStr;
 use num::rational::Ratio;
 use num::{BigUint, CheckedSub, Integer};
 use num_derive::{Num, NumOps, One, Zero};
+use num_traits::Pow;
 use serde::{Deserialize, Deserializer, Serialize, Serializer};
 
+/// DAI has precision of 18 decimals
+/// For CPM we have 3 decimals precision, but that's for 1000 (3 decimals more)
+/// This in terms means we need 18 - (3 + 3) = 12 decimals precision
+pub const GLOBAL_MULTIPLIER: Multiplier = Multiplier(12);
```
Ivshti: why is this not 10**12 rather than 12? it would save you from having to do a pow operation

elpiel: Good idea! I will impl it.

primitives/src/big_num.rs
shirshak55: TryFrom would be better as 0 may be magic :( I think error should be handled properly

elpiel:
Closing this in favor of the v5 Unified Precision of (8) #384

Issue #270, opened 2020-02-27
Sentry Fixes:


samparsky:
- [x]  Access forbidden referrer
- [x]  /channel/validate route
- [x]  Request body validation for /channel,
- [x]  SerHex panics on missing input
- [x] Fee constraint voilation
- [x] Match JS error response

PR #271, opened 2020-02-27
Fix route discrepancies

samparsky:
https://github.com/AdExNetwork/adex-validator-stack-rust/issues/270


sentry/src/routes/channel.rs
```diff
@@ -43,22 +45,32 @@ pub async fn create_channel<A: Adapter>(
 ) -> Result<Response<Body>, ResponseError> {
     let body = hyper::body::to_bytes(req.into_body()).await?;
 
-    let channel = serde_json::from_slice::<Channel>(&body)?;
+    let channel = serde_json::from_slice::<Channel>(&body)
+        .map_err(|e| ResponseError::FailedValidation(e.to_string()))?;
```
elpiel: Wait what?! Why is deserialization a FailedValidation, but `validate_channel` result is only BadRequest?

sentry/src/event_aggregator.rs
```diff
@@ -125,25 +126,35 @@ impl EventAggregator {
...
         let has_access = check_access(
             &app.redis,
-            &session,
+            session,
             &app.config.ip_rate_limit,
             &record.channel,
             events,
         )
         .await;
         if let Err(e) = has_access {
```
elpiel: you can simply use `map_err`

sentry/src/event_aggregator.rs
```diff
@@ -125,25 +126,35 @@ impl EventAggregator {
...
         }
 
         events
             .iter()
             .for_each(|ev| event_reducer::reduce(&record.channel, &mut record.aggregate, ev));
 
-        if ANALYTICS_RECORDER.is_some() {
+        // only time we don't have session is during
+        // an unauthenticated close event
+        if ANALYTICS_RECORDER.is_some() && session.is_some() {
```
elpiel: please `match`, we will remove the `expect` and it will look more clean.

sentry/src/access.rs
```diff
@@ -166,6 +180,29 @@ async fn apply_rule(
     }
 }
 
+fn forbidden_referrer(session: &Session) -> bool {
+    if let Some(hostname) = session
```
elpiel: please use `match`

sentry/src/access.rs
```diff
@@ -218,7 +255,7 @@ mod test {
     async fn session_uid_rate_limit() {
         let (config, redis) = setup().await;
 
-        let session = Session {
+        let session = &Session {
```
elpiel: You can simply use `Some(&session)` instead of creating `&Session`

primitives/src/validator.rs
```diff
@@ -16,7 +16,25 @@ pub enum ValidatorError {
...
+    [u8; 20],
+);
+
+fn validator_id_from_str<'de, D>(deserializer: D) -> Result<[u8; 20], D::Error>
+where
+    D: Deserializer<'de>,
+{
+    let validator_id = String::deserialize(deserializer)?;
+    if validator_id.is_empty() || validator_id.len() != 42 {
+        return Err(serde::de::Error::custom("invalid validator id".to_string()));
```
elpiel: This error is a bit vague, because on invalid hex string it is still an `invalid validator id`

primitives/src/validator.rs
```diff
@@ -67,6 +85,12 @@ impl TryFrom<&str> for ValidatorId {
             value
         };
 
+        if hex_value.len() != 40 {
```
elpiel: please use match with the other case above, instead of splitting it on multiple checks/lines.
Also I think that more explicit errors will help even us when debugging errors.

sentry/src/routes/channel.rs
```diff
@@ -169,7 +192,7 @@ pub async fn insert_events<A: Adapter + 'static>(
     let session = req
         .extensions()
         .get::<Session>()
-        .expect("request session")
+        .map(ToOwned::to_owned)
```
elpiel: You already receive a reference from `get()`, why do you need to own it and then `get_ref` again?
Just `get()` and then use it for the `record()`

primitives/src/validator.rs
samparsky: I am not sure what you mean by vague, it's a validator id param and if it's not a valid hex then it's an invalid validator id

elpiel: For debugging purposes it would be nice to distinguish between `empty/invalid length` and invalid `hex` string that you will get from `FromHex::from_hex`

primitives/src/validator.rs
samparsky: `match` wouldn't be cleaner in this scenario, and joining the cases together would make it further confusing

sentry/src/access.rs
samparsky: This is due to using `session` in multiple places having `Some(&session)` it would be impossible has it would be moved & would require a clone

sentry/src/event_aggregator.rs
samparsky: we can't use match has the match arms would be different

sentry/src/routes/channel.rs
samparsky: yeah, the difference is in the responses sent to the request client

sentry/src/routes/channel.rs
samparsky: This is because we use `req.into_body` which consumes `req` and invalidates the references from `get()`

primitives/src/validator.rs
samparsky: although I have changed the error message to make it more descriptive

primitives/src/validator.rs
samparsky: yeah i changed the error message to invalid length

sentry/src/routes/channel.rs
elpiel: It's a bit confusing reading the code like this, would be good to change it to remove confusion.

samparsky: No its not, `validate_channel` is not `FailedValidation` because its not checking the request body. `FailedValidation` is the error response  for checking  the request body

samparsky: BadRequest is for when you pass in valid request body but during processing the request something is wrong

PR #273, opened 2020-02-28
Validator worker timeout errors

elpiel:


validator_worker/src/main.rs
```diff
@@ -202,23 +202,25 @@ async fn validator_tick<A: Adapter + 'static>(
...
-            _ => (),
-        },
-        SpecValidator::None => {
-            return Err(ValidatorWorkerError::Failed(
-                "validatorTick: processing a channel where we are not validating".to_string(),
-            ))
-        }
-    };
-    Ok(())
+        SpecValidator::Leader(_) => timeout(duration, leader::tick(&sentry))
```
samparsky: yeah, I am not sure this better than the previous implementation. With the previous implementation it's very easy to understand that a Result is embedded in another Result but with this refactor it's not clear at all

elpiel: Moved it back to `match`

PR #277, opened 2020-03-02
Issue #260 preserve original error message

### LINKED ISSUE #260: preserve original error message in adapters/ethereum
the ethereum adapter erases the original error messages by mapping them to messages like "Invalid keystore location provided" or "Failed to create account"

The problem with that is that the messages are untuitive, but also that they erase information

E.g. failing to read a file may be due to a non-existant file, or it could be a permission denied, or it could even be a "too many open files" error; but we just show "Invalid keystore location provided", which is unclear and even inaccurate for some of those scenarios

same goes for "Failed to create account", which happens when your passphrase is incorrect; and also for other messages

**Solution:** wrap the original error into an enum and print the enum with it's contents

elpiel:
Fixes #260 #278 #280


adapter/src/dummy.rs
```diff
@@ -30,16 +33,28 @@ impl DummyAdapter {
...
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "Dummy Adapter error occurred!")
+    }
+}
+
+impl AdapterErrorKind for Error {}
+
 impl Adapter for DummyAdapter {
-    fn unlock(&mut self) -> AdapterResult<()> {
+    type AdapterError = Error;
```
samparsky: We don't need different errors for Ethereum/Dummy Adapter

samparsky: For our use case sharing an Error struct between `Ethereum/Dummy` Adapter works fine. Also, if we add additional adapters in the future easily reusable and extensible

adapter/src/ethereum.rs
```diff
@@ -113,95 +110,103 @@ impl Adapter for EthereumAdapter {
...
-    fn verify(&self, signer: &ValidatorId, state_root: &str, sig: &str) -> AdapterResult<bool> {
+    fn verify(
+        &self,
+        signer: &ValidatorId,
+        state_root: &str,
+        sig: &str,
+    ) -> AdapterResult<bool, Self::AdapterError> {
         if !sig.starts_with("0x") {
-            return Err(AdapterError::Signature("not 0x prefixed hex".to_string()));
+            return Err(StateRootError::SignatureNotPrefixed.into());
```
samparsky: Signature errors aren't a type of StateRootError it's quite different. Signature errors could either standalone or may be part of some struct with a more suitable name context

adapter/src/ethereum.rs
```diff
@@ -422,6 +431,182 @@ pub fn ewt_verify(
...
+    impl std::error::Error for Error {}
+
+    impl AdapterErrorKind for Error {}
+
+    impl fmt::Display for Error {
+        fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+            use Error::*;
+
+            match self {
+                Keystore(err) => write!(f, "Keystore error - {}", err),
```
samparsky: This format isn't great for parsing error logs. We should choose an error format  that all errors should adhere to

I suggest the below:

```
Title: Description : error  
```
e.g.

```
PublicKeyRecovery : Recovering the public key from the signature : {}
```

or

```
PublicKeyRecovery :  {err}
```

primitives/src/adapter.rs
```diff
@@ -1,45 +1,52 @@
...
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         match self {
-            AdapterError::Authentication(error) => write!(f, "Authentication error: {}", error),
-            AdapterError::EwtVerifyFailed(error) => write!(f, "Ewt verification error: {}", error),
-            AdapterError::Authorization(error) => write!(f, "Authorization error: {}", error),
-            AdapterError::Configuration(error) => write!(f, "Configuration error: {}", error),
-            AdapterError::Signature(error) => write!(f, "Signature error: {}", error),
-            AdapterError::InvalidChannel(error) => write!(f, "{}", error),
-            AdapterError::Failed(error) => write!(f, "error: {}", error),
+            Error::Authentication(error) => write!(f, "Authentication error: {}", error),
```
samparsky: Do we really need the `error` word here? But if included  it should be of format camelcase
```
AuthenticationError: 
```

validator_worker/src/error.rs
```diff
@@ -1,25 +1,47 @@
...
 
-impl Display for ValidatorWorker {
-    fn fmt(&self, f: &mut Formatter<'_>) -> Result {
         match self {
-            ValidatorWorker::Configuration(err) => write!(f, "Configuration error: {}", err),
-            ValidatorWorker::Failed(err) => write!(f, "error: {}", err),
-            ValidatorWorker::Channel(channel_id, err) => {
-                write!(f, "Channel {}: {}", channel_id, err)
-            }
+            SentryApi(err) => write!(f, "Sentry Api error - {}", err),
```
samparsky: Again here we should prefer camelcase error titles and also is the "description" really needed?

adapter/src/dummy.rs
```diff
@@ -1,9 +1,12 @@
 use futures::future::{BoxFuture, FutureExt};
-use primitives::adapter::{Adapter, AdapterError, AdapterResult, DummyAdapterOptions, Session};
+use primitives::adapter::{
+    Adapter, AdapterErrorKind, AdapterResult, DummyAdapterOptions, Error as AdapterError, Session,
+};
```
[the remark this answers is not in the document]
elpiel: They are in their own modules. the `adapter` crate has own `Error` and same goes for `dummy` and `ethereum`

adapter/src/dummy.rs
elpiel: Not every error apply for every adapter. Dummy adapter doesn't use Keystore, why should you have a Keystore error? It's not a general error.

primitives/src/adapter.rs
elpiel: Yep, we could remove them and leave only relevant context, as I did with the new errors. The PR got a bit bigger though so I stopped

validator_worker/src/error.rs
elpiel: Otherwise you won't see the context of this error when the error is logged/printed. You will only see the underlying error which can lead to same error (`adapter::Error`) having the same error message in 2 different cases for the child error.

validator_worker/src/sentry_interface.rs
```diff
@@ -70,28 +112,36 @@ impl<T: Adapter + 'static> SentryApi<T> {
...
+    // @TODO: Remove logging & fix `try_join_all` @see: https://github.com/AdExNetwork/adex-validator-stack-rust/issues/278
     pub async fn propagate(&self, messages: &[&MessageTypes]) {
         let channel_id = format!("0x{}", hex::encode(&self.channel.id));
         if let Err(e) = try_join_all(self.propagate_to.iter().map(|(validator, auth_token)| {
             propagate_to(&channel_id, &auth_token, &self.client, &validator, messages)
         }))
         .await
         {
-            error!(&self.logger, "Propagation error: {}", e; "module" => "sentry_interface", "in" => "SentryApi");
+            error!(&self.logger, "Propagation error - {}", e; "module" => "sentry_interface", "in" => "SentryApi");
```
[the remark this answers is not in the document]
elpiel: will change, I thought though that having `Some Error: Another error: Third error: fourth error with some values: channel: 0x...` can be a bit confusing.

adapter/src/ethereum.rs
elpiel: Yes, we need to find a different name for this error, but in a nutshell we want to specify that that decoding the `signature`, `state_root` or verifying the `address, signature, message` failed. Another way is to completely remov this error by accepting something like `Vec<u8>/[u8]`.
This can allow us to still have `String/&str` for DummyAdapter & remove all the `decode`, `starts_with("0x")`, etc.

validator_worker/src/sentry_interface.rs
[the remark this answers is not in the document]
samparsky: It's very easy to write a parser for it that way if we end up wanting to do error analytics

adapter/src/dummy.rs
samparsky: You are thinking in terms Dummy/Ethereum Adapter.  The DummyAdapter is just for development (it doesn't warrant the added complexity), if possible it shouldn't be compiled for release. When/If we add another adapter it's most likely going to use the same errors with the Ethereum Adapter.

elpiel: I think in terms of `trait Adapter` you cannot generalize every single error an implementation of the trait can have in a single `Error` since every trait implementation can have different impl Adapter specific errors.

elpiel: There might be other ways to do it. Think of std::io::Error for example - you have the Error, but also the ErrorKind. I made it slightly different - leaving the non-specific Adapter Errors in the `adapter::Error` and then having a trait `AdapterErrorKind` to be implemented by each adapter specific Error.
_This can be changed later on._

https://doc.rust-lang.org/std/io/struct.Error.html

samparsky: There is no issue with AdapterErrorKind.

The point is Dummy Adapter throws every error Ethereum Adapter has. The only reason why it's not visible is because of the oversimplification of signatures to formatted strings and ignoring a number of cases because "development" etc. 

There is no point in having separate error struct for both of them. They both should throw the same errors. 

The Ethereum Adapter Error struct should also be used for Dummy Adapter.  In fact, we can model `Keystore error` for `DummyAdapter`  as a case where identity passed doesn't match the identity in `prep-db` because "technically" that is it  Keystore. 

The error struct should be shared, different error structs doesn't really offer any benefits. 
It leads having to duplicate the same errors in ETH to Dummy. That's why in the code for Dummy Adapter https://github.com/AdExNetwork/adex-validator-stack-rust/blob/cfc8c6d73b25158e8b3cb31aa20b49b0d160bba7/adapter/src/dummy.rs#L40 there is no implementation there because it would require duplicating error types.

adapter/src/ethereum.rs
elpiel: Since the errors are nested you actually get something like:
`Mar 11 12:58:16.198 ERRO sentry: Adapter specific: Keystore: "address" is invalid: Invalid argument Keystore ValidatorId, module: middleware-auth`
When the error bubbles up to the logging and starts composing the error message, it will get all those nested errors and nest them as `Error msg 1: Error msg 2: Error msg 3: Error msg 4: Error msg 5`

I did however removed the `error` part and replaced the `-` with `:`.

elpiel:
ping @samparsky

adapter/src/ethereum.rs
```diff
@@ -422,6 +431,182 @@ pub fn ewt_verify(
     Ok(verified_payload)
 }
 
+mod error {
```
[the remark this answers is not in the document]
samparsky: Let's move this to a separate file

validator_worker/src/follower.rs
```diff
@@ -1,28 +1,44 @@
...
 }
 
-enum NewStateResult {
-    Ok,
-    Err(InvalidNewState),
+#[derive(Debug)]
+pub enum ApproveStateResult<AE: AdapterErrorKind> {
+    Sent(Vec<PropagationResult<AE>>),
+    /// Conditions for handling the new state haven't been met
+    NotSent,
```
samparsky: We don't need `NotSent`

validator_worker/src/producer.rs
```diff
@@ -2,17 +2,28 @@ use std::error::Error;
...
+#[derive(Debug)]
+pub enum TickStatus<AE: AdapterErrorKind> {
+    AccountingSent {
+        channel: ChannelId,
+        balances: BalancesMap,
+        new_accounting: Accounting,
+        accounting_propagation: Vec<PropagationResult<AE>>,
+        event_counts: usize,
+    },
+    AccountingNotSent(BalancesMap),
```
samparsky: 1. Since this is `producer` specific we can remove the `Accounting` prefixes
2. `AccountingNotSent` relays another meaning, a better descriptive name would be `NoNewEventAggr` or `ZeroEvAggr`
3. For `AccountingSent` we don't need `BalancesMap`, `Accounting`.  `Channel` & `accounting_propagation`, `event_counts` and `state_root`  would be great

validator_worker/src/sentry_interface.rs
```diff
@@ -26,23 +28,70 @@ pub struct SentryApi<T: Adapter> {
...
+
+impl<AE: AdapterErrorKind> fmt::Display for Error<AE> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        use Error::*;
+
+        match self {
+            BuildingClient(err) => write!(f, "Building client: {}", err),
+            Request(err) => write!(f, "Making a request: {}", err),
+            ValidatorAuthentication(err) => {
+                write!(f, "Getting authentication for validator: {}", err)
```
samparsky: We should make these simpler,  this is quite okay
```js
BuildingClient(err) => write!(f, "BuildingClient: {}", err),
            Request(err) => write!(f, "Request: {}", err),
            ValidatorAuthentication(err) => {
                write!(f, "ValidatorAuthentication: {}", err)
```

validator_worker/src/follower.rs
samparsky: It should be an `Option<Vec<PropagationResult<AE>>>`

validator_worker/src/follower.rs
```diff
@@ -132,5 +159,8 @@ async fn on_error<'a, A: Adapter + 'static>(
         })])
         .await;
 
-    NewStateResult::Err(status)
+    ApproveStateResult::Err {
```
samparsky: This should be named `ApproveStateResult::RejectState`

validator_worker/src/follower.rs
```diff
@@ -1,28 +1,44 @@
...
-enum NewStateResult {
-    Ok,
-    Err(InvalidNewState),
+#[derive(Debug)]
+pub enum ApproveStateResult<AE: AdapterErrorKind> {
+    Sent(Vec<PropagationResult<AE>>),
+    /// Conditions for handling the new state haven't been met
+    NotSent,
+    Err {
+        new_state: InvalidNewState,
```
samparsky: This should be `reason` and not `new_state`

validator_worker/src/sentry_interface.rs
elpiel: This is only for Display (i.e. printing out) we use the Debug impl for the logging so this should be fine for reading

validator_worker/src/producer.rs
elpiel: We do actually need the `BalancesMap`, we use it in `follower::tick` and the `new_accounting` in `leader::tick`

validator_worker/src/follower.rs
elpiel: done

elpiel:
ping @samparsky

Issue #280, opened 2020-03-12
Log results from leader/follower ticks


elpiel:
* propagation of Heartbeat
* propagation of NewState (leader) /ApproveState (follower)
* Errors returned from the ticks

Introduced in #278

PR #281, opened 2020-03-13
Issue #278 Propagation logging

### LINKED ISSUE #278: SentryApi propagation logging and `try_join_all` fix
- [x] Move logging out of `SentryApi.propagate()` #281 
- [x] Fix `try_join_all` #281 

1. We can return a more complete error that can indicate exactly which validator request (Channel id, Validator Id, etc) failed and why (providing the `reqwest::Error`)
2. `try_join_all` will fail on the first error, which means that if the Follower errors for a reason, the Leader might not get the `message` as `try_join_all` can return the error and cancel the other futures (see https://docs.rs/futures/0.3.4/futures/future/fn.try_join_all.html)

https://github.com/AdExNetwork/adex-validator-stack-rust/blob/e8dd5c07ec/validator_worker/src/sentry_interface.rs#L81

elpiel:
Fixes #278 #280

### DIFF
--- a/sentry/src/middleware/auth.rs
+++ b/sentry/src/middleware/auth.rs
@@ -83,7 +83,7 @@ fn get_request_ip(req: &Request<Body>) -> Option<String> {
         .get("true-client-ip")
         .or_else(|| req.headers().get("x-forwarded-for"))
         .and_then(|hv| hv.to_str().map(ToString::to_string).ok())
-        .map(|token| token.split(',').nth(0).map(ToString::to_string))
+        .map(|token| token.split(',').next().map(ToString::to_string))
         .flatten()
 }
 
--- a/validator_worker/src/main.rs
+++ b/validator_worker/src/main.rs
@@ -202,6 +202,7 @@ async fn validator_tick<A: Adapter + 'static>(
         .map_err(ValidatorWorkerError::SentryApi)?;
     let duration = Duration::from_millis(config.validator_tick_timeout as u64);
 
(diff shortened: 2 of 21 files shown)


sentry/src/middleware/auth.rs
```diff
@@ -83,7 +83,7 @@ fn get_request_ip(req: &Request<Body>) -> Option<String> {
         .get("true-client-ip")
         .or_else(|| req.headers().get("x-forwarded-for"))
         .and_then(|hv| hv.to_str().map(ToString::to_string).ok())
-        .map(|token| token.split(',').nth(0).map(ToString::to_string))
+        .map(|token| token.split(',').next().map(ToString::to_string))
```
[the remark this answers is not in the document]
elpiel: `Clippy` - apparently it's an error in the new clippy

validator_worker/src/main.rs
```diff
@@ -195,15 +196,16 @@ async fn validator_tick<A: Adapter + 'static>(
     channel: Channel,
     config: &Config,
     logger: &Logger,
-) -> Result<(), ValidatorWorkerError<A::AdapterError>> {
+) -> Result<(ChannelId, Box<dyn Debug>), ValidatorWorkerError<A::AdapterError>> {
```
samparsky: why is it validator_tick returning `(ChannelId, Box<dyn Debug>)` if we don't need / use it?

elpiel: It's logged

PR #304, opened 2020-06-03
Math functions from AIP#31

### LINKED ISSUE #302: Targeting DSL Math Functions
Implement the math functions `mul`, `mod`, `add`, `sub`,  `min` and `max` as part of https://github.com/AdExNetwork/adex-validator-stack-rust/issues/296

simzzz:
fixes #302

### DIFF
--- a/primitives/src/targeting/eval.rs
+++ b/primitives/src/targeting/eval.rs
@@ -1,7 +1,12 @@
 use crate::BigNum;
 use serde::{Deserialize, Serialize};
 use serde_json::{value::Value as SerdeValue, Number};
-use std::{convert::TryFrom, fmt, ops::{Div, Mul, Rem, Add, Sub}, str::FromStr};
+use std::{
+    convert::TryFrom,
+    fmt,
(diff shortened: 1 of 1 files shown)


primitives/src/targeting/eval.rs
```diff
@@ -193,17 +204,34 @@ fn eval(input: &Input, output: &mut Output, rule: &Rule) -> Result<Option<Value>
...
+            )?
+            .ok_or(Error::TypeError)?;
+            let product_rule = Rule::Value(product);
+            let boxed_rule = Box::new(product_rule);
+            let value = eval(
+                input,
+                output,
+                &Rule::Function(Function::Div(boxed_rule, third_rule.clone())),
+            )?
+            .ok_or(Error::TypeError)?;
```
elpiel, line 221: loose the `ok_or(...)` and directly return the value. It is already an `Option` and `ok_or` just maps the `None` => `Error::TypeError` in this case, but we don't need it.

Issue #296, opened 2020-05-04
AIP#31 Targeting rules engine


elpiel:
# https://github.com/AdExNetwork/aips/issues/31
Part of AdExNetwork/adex-supermarket#9 

## Examples

https://github.com/AdExNetwork/aips/blob/master/examples/aip31.js

## Rust `eval` signature

```rust
enum Error {
    TypeError,
    UnknownVariable
}

fn eval(input: &Input, output: &mut Output, rule: &Rule) -> Result<Option<Value>, Error>
```

See #297 for the impl of Input/Output variables of https://github.com/AdExNetwork/aips/issues/31

### Values:
- [x] #297 `Bool(bool)`
- [x] #297 `Number(Number)`
- [x] #297 `String(String)`
- [x] #297 `Array(Vec<Value>)`
- [x] #297 `BigNum(BigNum)`

#### Math
> If a math function is invoked with numeric, but different types, it will try to cast all the arguments to a BigNumber
- [x] #297 `Div(Box<Rule>, Box<Rule>)`
  - [x] TODO: Make the casting change according to the AIP#31 new changes
- [x] Mul
- [x] Mod
- [x] Add
- [x] Sub
- [x] Max
- [x] Min

#### Variables
- [x] #297 `Get(String)`
  - [x] PR #325  `Get` should be applied to `Output` as well
- [x] #297 `Set(String, Box<Rule>)`

       > NOTE: rules that use variables that are only accessible on the AdView cannot change the price, since it's ultimately determined by the validator.

#### Syntax sugar
- [x] onlyShowIf
- [x] ~? setMul **`*=`**~ Not impled in JS or Rust atm
- [x] ~? setAdd **`+=`**~ Not impled in JS or Rust atm

#### Price helpers
- [x] #307 (PR #305 & PR #321 ) getPriceInUSD @simzzz 

  > getPriceInUSD will return the currently set price for a specific event, approximated in USD; this is useful for the ad slot rules, as it allows setting min CPM regardless of asset type

#### Logic
- [x] or
- [x] #297 `And(Box<Rule>, Box<Rule>)`
- [x] xor
- [x] not

#### Flow control
- [x] #297 `If(Box<Rule>, Box<Rule>)`
- [x] ifNot
- [x] ifElse
- [x] PR #305 do

#### Comparison
- [x] lt
- [x] lte
- [x] gt
- [x] gte
- [x] eq
- [x] neq
- [x] nin
- [x] in
[...]

PR #305, opened 2020-06-08
The rest of the DSL functions

### LINKED ISSUE #307: AIP#31 impl `getPriceInUsd` DSL function
@Ivshti :
Just convert the given `BN` to a `float` (`f64`) of dollars
Depending on the `depositAsset` of the `Channel`
Just hardcode the most popular stablecoins and their multipliers
Eg if the depositAsset is:
-  DAI: `BN / 10**18`
- Tether: `BN / 10**6`

# Tasks
- [x] #305 Configure the most popular stablecoins and their multipliers (not necessary in `Config`, can be a `static`/`const` list)
- [x] PR #321  impl `getPriceInUSD` returning a `f64` number

simzzz:
Part of #296 
closes #307


primitives/src/targeting/eval.rs
```diff
@@ -401,6 +518,81 @@ fn eval(input: &Input, output: &mut Output, rule: &Rule) -> Result<Option<Value>
...
+            Some(Value::Bool((a && !b) || (!a && b)))
+        }
+        Function::Not(first_rule) => {
+            let a = eval(input, output, first_rule)?
+                .ok_or(Error::TypeError)?
+                .try_bool()?;
+
+            Some(Value::Bool(!a))
+        }
+        Function::Lt(first_rule, second_rule) => {
```
elpiel, line 548: these functions should work the same way as Math functions.
We've discussed it, this comment is just a reminder.

primitives/src/targeting/eval.rs
```diff
@@ -401,6 +518,81 @@ fn eval(input: &Input, output: &mut Output, rule: &Rule) -> Result<Option<Value>
...
+        }
+        Function::Gte(first_rule, second_rule) => {
+            let first_eval = first_rule.eval(input, output)?.ok_or(Error::TypeError)?;
+            let second_eval = second_rule.eval(input, output)?.ok_or(Error::TypeError)?;
+
+            let lhs = BigNum::try_from(first_eval)?;
+            let rhs = BigNum::try_from(second_eval)?;
+            Some(Value::Bool(lhs.ge(&rhs)))
+        }
+        Function::Equals(first_rule, second_rule) => {
```
elpiel, line 580: This should work with **any** value. Not just numbers

primitives/src/targeting/eval.rs
```diff
@@ -411,6 +603,94 @@ fn eval(input: &Input, output: &mut Output, rule: &Rule) -> Result<Option<Value>
...
+                .ok_or(Error::TypeError)?
+                .try_array()?;
+            let second_eval = second_rule
+                .eval(input, output)?
+                .ok_or(Error::TypeError)?
+                .try_number()?
+                .as_u64()
+                .ok_or(Error::TypeError)?;
+            let index = second_eval as usize;
+            let value = &first_eval[index];
```
elpiel, line 645: This can panic!
use `remove_index` after checking if the `index` exist OR `get()` with `clone()` (for now)
https://stackoverflow.com/a/45804069

primitives/src/targeting/eval.rs
```diff
@@ -411,6 +603,94 @@ fn eval(input: &Input, output: &mut Output, rule: &Rule) -> Result<Option<Value>
...
+                .eval(input, output)?
+                .ok_or(Error::TypeError)?
+                .try_array()?;
+            let second_eval = second_rule
+                .eval(input, output)?
+                .ok_or(Error::TypeError)?
+                .try_number()?
+                .as_u64()
+                .ok_or(Error::TypeError)?;
+            let index = second_eval as usize;
```
elpiel, line 644: This can panic! Please use the `TryFrom` https://doc.rust-lang.org/std/primitive.usize.html#impl-TryFrom%3Cu64%3E

primitives/src/targeting/eval.rs
```diff
@@ -411,6 +603,94 @@ fn eval(input: &Input, output: &mut Output, rule: &Rule) -> Result<Option<Value>
...
+            let first_eval = first_rule
+                .eval(input, output)?
+                .ok_or(Error::TypeError)?
+                .try_string()?;
+            let second_eval = second_rule
+                .eval(input, output)?
+                .ok_or(Error::TypeError)?
+                .try_string()?;
+
+            let after_split: Vec<&str> = first_eval.split(&second_eval).collect();
```
elpiel, line 658: Instead of collecting and then creating an iterator again, just map the values before the `collect()`

primitives/src/targeting/eval.rs
```diff
@@ -411,6 +722,102 @@ fn eval(input: &Input, output: &mut Output, rule: &Rule) -> Result<Option<Value>
...
+        Function::Between(first_rule, second_rule, third_rule) => {
+            let first_eval = first_rule.eval(input, output)?.ok_or(Error::TypeError)?;
+            let second_eval = second_rule.eval(input, output)?.ok_or(Error::TypeError)?;
+            let third_eval = third_rule.eval(input, output)?.ok_or(Error::TypeError)?;
+
+            let value = BigNum::try_from(first_eval)?;
+            let start = BigNum::try_from(second_eval)?;
+            let end = BigNum::try_from(third_eval)?;
+
+            Some(Value::Bool(value.ge(&start) && value.le(&end)))
```
simzzz, line 742: I assumed the second value will always be smaller than the third value

primitives/src/targeting/eval.rs
```diff
@@ -411,6 +722,102 @@ fn eval(input: &Input, output: &mut Output, rule: &Rule) -> Result<Option<Value>
...
+                .try_number()?
+                .as_u64()
+                .ok_or(Error::TypeError)?;
+            let index = usize::try_from(second_eval).unwrap();
+            let value = if first_eval.get(index).is_none() {
+                None
+            } else {
+                Some(first_eval.swap_remove(index))
+            };
+            let value = value.ok_or(Error::TypeError)?;
```
simzzz, line 761: Maybe add a more descriptive error here?

primitives/src/targeting/eval.rs
elpiel: even if it's not then 1 of the 2 checks will fail. I.e. if third value (end) is smaller than second (start) then either `value.ge(&strart)` will fail or `value.le(&end)` as `value` can't satisfy both at the same time.

primitives/src/targeting/eval.rs
elpiel: Not really needed at this point.
However you can simply leave the if-else at the end and drop the:

```rust
let value = value.ok_or(Error::TypeError)?;
Some(value)
```
with smth like:
```rust
if first_eval.get(index).is_none() {
                return Err(Error::TypeError);
            } else {
                Some(first_eval.swap_remove(index))
            }
```

primitives/src/targeting/eval.rs
```diff
@@ -411,6 +714,107 @@ fn eval(input: &Input, output: &mut Output, rule: &Rule) -> Result<Option<Value>
...
+                .eval(input, output)?
+                .ok_or(Error::TypeError)?
+                .try_array()?;
+            let second_eval = second_rule
+                .eval(input, output)?
+                .ok_or(Error::TypeError)?
+                .try_number()?
+                .as_u64()
+                .ok_or(Error::TypeError)?;
+            let index = usize::try_from(second_eval).unwrap();
```
elpiel, line 755: do not `unwrap()`, this will `panic!` if it fails. `map_err` to an `Error`

primitives/src/targeting/eval.rs
```diff
@@ -411,6 +714,107 @@ fn eval(input: &Input, output: &mut Output, rule: &Rule) -> Result<Option<Value>
...
+        }
+        Function::NotIn(first_rule, second_rule) => {
+            let a = eval(input, output, first_rule)?.ok_or(Error::TypeError)?;
+            let b = eval(input, output, second_rule)?
+                .ok_or(Error::TypeError)?
+                .try_array()?;
+
+            Some(Value::Bool(!b.contains(&a)))
+        }
+        Function::Between(first_rule, second_rule, third_rule) => {
```
elpiel, line 733: This should work the same way as `mulDiv` handling all numbers. Also a combination of rules can be `eval`-ed instead of writing the logic.

primitives/src/targeting/eval.rs
```diff
@@ -97,8 +97,29 @@ pub enum Function {
...
+    IfElse(Box<Rule>, Box<Rule>, Box<Rule>),
     And(Box<Rule>, Box<Rule>),
+    Or(Box<Rule>, Box<Rule>),
+    Xor(Box<Rule>, Box<Rule>),
+    Not(Box<Rule>),
+    Lt(Box<Rule>, Box<Rule>),
+    Lte(Box<Rule>, Box<Rule>),
+    Gt(Box<Rule>, Box<Rule>),
+    Gte(Box<Rule>, Box<Rule>),
+    Equals(Box<Rule>, Box<Rule>),
```
elpiel, line 110: Keep in mind that this will become `equals: {...}` and not the DSL `eq`.
This is due to `#[serde(rename_all = "camelCase")]`, by default it uses the enum variant as a tag (see https://serde.rs/enum-representations.html).
This also includes `NotEquals`, `NotIn`.
You can use (see https://serde.rs/attributes.html):
```rust
#[derive(Serialize, Deserialize)]
#[serde(rename = "e")]  // <-- this is also a container attribute
enum E {
    #[serde(rename = "a")]  // <-- this is a variant attribute
    A(String),
}
```

however it's maybe better to Document (use doc comments to explain what is `eq`, `nin`, etc.) which type is what and keep the names the same as the DSL names.

primitives/src/targeting/eval.rs
```diff
@@ -401,6 +519,191 @@ fn eval(input: &Input, output: &mut Output, rule: &Rule) -> Result<Option<Value>
...
+        }
+        Function::Xor(first_rule, second_rule) => {
+            let a = eval(input, output, first_rule)?
+                .ok_or(Error::TypeError)?
+                .try_bool()?;
+            let b = eval(input, output, second_rule)?
+                .ok_or(Error::TypeError)?
+                .try_bool()?;
+
+            Some(Value::Bool((a && !b) || (!a && b)))
```
elpiel, line 540: Forgot about this last time, but you can use the XOR (`^`) operator instead, see:
https://doc.rust-lang.org/std/primitive.bool.html#impl-BitXor%3Cbool%3E

primitives/src/targeting/eval.rs
elpiel: This still hasn't been resolved. The reason is that will impact the performance (collecting twice will result in memory allocation twice)

```rust
let after_split = first_eval.split(&second_eval).map(Value::new_string).collect();
Some(Value::Array(after_split))
```

primitives/src/targeting/eval.rs
```diff
@@ -475,6 +879,61 @@ enum MathOperator {
     Min,
 }
 
+enum ComparisonOperator {
```
elpiel, line 882: can you document them or use full names for all?

primitives/src/targeting/eval.rs
```diff
@@ -475,6 +879,61 @@ enum MathOperator {
...
+}
+
+fn handle_comparisons<T: PartialOrd>(
+    lhs: T,
+    rhs: T,
+    ops: ComparisonOperator,
+) -> Result<bool, Error> {
+    match ops {
+        ComparisonOperator::Lt => {
+            let is_lt = lhs < rhs;
```
elpiel, line 911: Note: you could actually invoke the functions of PartialEq/PartialOrd directly.
I.e. `lhs.lt(rhs)`

primitives/src/targeting/eval.rs
```diff
@@ -475,6 +879,61 @@ enum MathOperator {
...
+            },
+        },
+    }
+}
+
+fn handle_comparisons<T: PartialOrd>(
+    lhs: T,
+    rhs: T,
+    ops: ComparisonOperator,
+) -> Result<bool, Error> {
```
elpiel, line 908: this function doesn't fail in any case, is there a reason for the `Result`?

primitives/src/targeting/eval_test.rs
```diff
@@ -0,0 +1,1244 @@
...
+            boost: 1.0,
+            price: Default::default(),
+        };
+        let result = Value::BigNum(200.into());
+        let rule = Rule::Function(Function::new_add(
+            Value::BigNum(100.into()),
+            Value::BigNum(100.into()),
+        ));
+        let rule_do = Rule::Function(Function::new_do(rule));
+        assert_eq!(Ok(Some(result.clone())), rule_do.eval(&input, &mut output));
```
elpiel, line 1021: No need for `.clone()`.

primitives/src/targeting/eval_test.rs
```diff
@@ -0,0 +1,1244 @@
...
+                Value::Array(vec![
+                    Value::BigNum(1.into()),
+                    Value::BigNum(2.into()),
+                    Value::BigNum(3.into()),
+                ]),
+                false,
+            ),
+        ];
+
+        for (value, arr, expected) in cases.iter() {
```
elpiel, line 1058: `vec!` + `into_iter()` and you can remove the `clone()`s and `*`

primitives/src/targeting/eval.rs
```diff
@@ -12,12 +12,46 @@ pub type Map = serde_json::value::Map<String, SerdeValue>;
...
+
 #[derive(Debug, Eq, PartialEq)]
 pub enum Error {
     TypeError,
     UnknownVariable,
 }
 
+fn get_deposit_asset_divisor(address: &String) -> Result<BigNum, Error> {
+    match address.as_str() {
+        "0x89d24A6b4CcB1B6fAA2625fE562bDD9a23260359" => Ok(BigNum::from(10u64.pow(18))),
```
elpiel, line 27: which one is which?

primitives/src/targeting.rs
```diff
@@ -102,11 +103,12 @@ pub struct Global {
     pub advertiser_id: String,
     pub country: String,
     pub event_type: String,
-    pub campiagn_id: String,
+    pub campaign_id: String,
     pub campaign_total_spent: String,
     pub campaign_seconds_active: u64,
     pub campaign_seconds_duration: u64,
     pub campaign_budget: BigNum,
+    pub deposit_asset: String,
```
elpiel, line 111: this can be removed once we merge #308 either here or in `dev` since the `global.channel` contains also the deposit asset:
https://github.com/AdExNetwork/adex-validator-stack-rust/blob/dev/primitives/src/channel.rs#L83

primitives/src/targeting/eval.rs
```diff
@@ -12,12 +12,46 @@ pub type Map = serde_json::value::Map<String, SerdeValue>;
...
+#[path = "eval_test.rs"]
+mod test;
+
 #[derive(Debug, Eq, PartialEq)]
 pub enum Error {
     TypeError,
     UnknownVariable,
 }
 
+fn get_deposit_asset_divisor(address: &String) -> Result<BigNum, Error> {
```
elpiel, line 25: `&String` should be `&str`, it's always better to have the slices (&str, &[Type]) instead of the owned values (&String, Vec<Type>).

primitives/src/targeting/eval.rs
```diff
@@ -12,12 +12,46 @@ pub type Map = serde_json::value::Map<String, SerdeValue>;
...
+    match address.as_str() {
```
elpiel, line 26: Could you check what should happen if the address is not found in the list? Maybe a `None` for the price would be better?!

We should probably add `USDC` as well. Maybe `DIA`, `USDT` & `USDC` is sufficient list?
 
@Ivshti maybe you could shine a light on the two questions?

primitives/src/targeting/eval_test.rs
```diff
@@ -0,0 +1,1346 @@
...
+                Rule::Value(string.clone()),
+                Rule::Value(starting.clone()),
+            ));
+            let expected = Some(Value::Bool(*expected));
+            assert_eq!(Ok(expected), rule.eval(&input, &mut output));
+        }
+    }
+
+    #[test]
+    fn test_get_dai_price_in_usd_eval() {
```
elpiel, line 1310: Could you extract the strings in `const`s (`DAI`, `USDC`, `USDT`) and we can keep the list in a `HashMap` as well and thus remove the need for the whole `fn get_deposit_asset_divisor()` and use the provided methods from the `HashMap` (you would need `lazy_static` for the `HashMap`).

This way the tests get redundant

primitives/src/targeting/eval.rs
```diff
@@ -887,7 +895,7 @@ fn eval(input: &Input, output: &mut Output, rule: &Rule) -> Result<Option<Value>
                 .try_bignum()?;
             let deposit_asset = &input.global.channel.deposit_asset;
 
-            let divisor = get_deposit_asset_divisor(&deposit_asset)?;
+            let divisor = DEPOSIT_ASSETS_MAP.get(deposit_asset).unwrap();
```
elpiel, line 898: No unwraps, please. Check what error we should return in this case (TypeError?) or maybe just value `0` for the amount

PR #341, opened 2020-10-26
Issue 337 align channellist response

### LINKED ISSUE #337: align Channel List response with JS impl
Add the 2 missing values from the response that are in the JS impl:

* `total`
* `page`

simzzz:
closes #337

### DIFF
--- a/sentry/src/routes/channel.rs
+++ b/sentry/src/routes/channel.rs
@@ -89,7 +89,6 @@ pub async fn channel_list<A: Adapter>(
         &query.creator,
         &query.validator,
         &query.valid_until_ge,
-        query.page,
     )
     .await?;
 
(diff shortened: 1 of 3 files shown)


sentry/src/routes/channel.rs
```diff
@@ -89,6 +89,7 @@ pub async fn channel_list<A: Adapter>(
         &query.creator,
         &query.validator,
         &query.valid_until_ge,
+        query.page,
```
samparsky, line 92: This could be `skip / limit` instead of passing  `query.page` since skip = limit * page.

samparsky: or we remove the skip param and do the calculation in `list_channels`

elpiel: Another option is to return `(total_pages, channels)` since the `db/channel` is handling DB requests and then construct the response here.

```
let (total_pages, channels) = list_channels(...)...;

Ok(ChannelListResponse {
            total_pages,
            total: total_pages,
            page,
            channels,
})
```

elpiel (approved):

PR #342, opened 2020-10-28
no heartbeats in response when query param is not provided

### LINKED ISSUE #339: `/channel/:id/last-approved` should not include `heartbeats` in response
This is a consistency issue between the JS impl & Rust impl.
When `GET /channel/:id/last-approved` is requested without `?withHeartbeat=true` the response should not include the `heartbeats` in the response.

Closed with #342

simzzz:
closes #339

### DIFF
--- a/adview-manager/src/lib.rs
+++ b/adview-manager/src/lib.rs
@@ -411,7 +411,6 @@ impl Manager {
             .collect::<Vec<_>>()
             .join("&");
 
-
         // Url adds a trailing `/`
         let url = format!(
             "{}units-for-slot/{}?pubPrefix={}&{}",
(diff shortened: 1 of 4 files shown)


elpiel (changes requested):
For this fix you just need to specify a `serde` attribute for the field.
Check the attribitues here: https://serde.rs/

Also please add a Documentation comment for `heartbeat` field to explain what the `Option` means, i.e.:
* `None` -> `withHeartbeat=true` wasn't passed
* `Some(vec![])` (empty vec) or `Some(heartbeats)` - `withHeartbeats=true` was passed

elpiel (approved):

elpiel:
LGTM!

PR #344, opened 2020-11-11
fromstr implemented for channelid

simzzz:


primitives/src/channel.rs
```diff
@@ -74,6 +75,18 @@ impl fmt::Display for ChannelId {
     }
 }
 
+impl FromStr for ChannelId {
```
elpiel, line 78: Could you unify the `channel_id_from_str` and this impl of `FromStr`?
The `channel_id_from_str` doesn't check for `0x` prefix which means they act differently.

Idk what will look ok given the serialization is handled differently, however a lot of the code could be removed if we do it.

elpiel (changes requested):

primitives/src/channel.rs
```diff
@@ -27,16 +27,21 @@ impl fmt::Debug for ChannelId {
...
-    if channel_id.is_empty() || channel_id.len() != 66 {
-        return Err(serde::de::Error::custom("invalid channel id".to_string()));
-    }
+    let channel_id = validate_channel_id(channel_id).map_err(serde::de::Error::custom)?;
+    <[u8; 32] as FromHex>::from_hex(channel_id).map_err(serde::de::Error::custom)
+}
 
-    <[u8; 32] as FromHex>::from_hex(&channel_id[2..]).map_err(serde::de::Error::custom)
+fn validate_channel_id(s: String) -> Result<String, FromHexError> {
+    if s.is_empty() || s.len() != 66 || &s[0..2] != "0x" {
```
elpiel, line 40: Use `&str` instead of `String`, borrow is better, since you avoid unnecessary clones (expensive copies)

Be careful when dealing with indices like `[0..2]`, they usually `panic!` if something goes wrong.
Use Non-panicing methods instead:

https://doc.rust-lang.org/std/primitive.str.html#method.get

They also work with ranges `get(2..)` and will return `None` (`is_empty()`) instead of panicking.
Just use a `match (s.get(0..2), s.get(2..))`, you can also match on the prefix `(Some(prefix), Some(hex)) if hex.len() == 64`.

Also you can make the return a [u8; 32]
`<[u8; 32] as FromHex>::from_hex(hex)`

primitives/src/channel.rs
```diff
@@ -79,11 +84,7 @@ impl FromStr for ChannelId {
     type Err = FromHexError;
 
     fn from_str(s: &str) -> Result<Self, Self::Err> {
-        if s.len() != 66 || &s[0..2] != "0x" {
-            Err(FromHexError::InvalidStringLength)
-        } else {
-            Self::from_hex(&s[2..])
-        }
+        ChannelId::from_hex(validate_channel_id(s.into())?)
```
elpiel, line 87: `s.into()` will clone the `&str` into a `String`, use the `&str` instead to avoid unnecessary cloning.

elpiel (changes requested):

elpiel (approved):

Issue #312, opened 2020-06-29
Remove struct TargetingTag


elpiel:
Since the introduction of [AIP#31](https://github.com/AdExNetwork/aips/issues/31) we no longer need it, so find and remove all places and mentions of it.
The `TargetingTag` is contained in it's own module so the whole module should also be removed.

**Note:** It requires changes to `adview-manager` crate as well

elpiel:
Closed with #345

PR #345, opened 2020-11-12
Issue #340 re-think targeting input

### LINKED ISSUE #340: Re-think Targeting Input AIP#31
There is a problem with the current `Input` for targeting. This is that while `Source` can serialize with proper `targeting::eval::Value`s, the `Map` (`HashMap<String, targeting::input::Value>`) can't since it's set by deserializing a Json value (`serde_json::Value`). This causes problems since using the `Map` as a base for e.g. `adview-manager` will result in loosing information about the values of the targeting input.

## Todos:

- [x] enable assertion of Targeting input in Supermarket - https://github.com/AdExNetwork/adex-supermarket/pull/29

https://github.com/AdExNetwork/adex-supermarket/blob/ca659b469da0bfe401e41f5fb47e27be0afeb3ad/src/units_for_slot_test.rs#L298-L299
- [x] Update Supermarket https://github.com/AdExNetwork/adex-supermarket/pull/29
- [x] Check Input variables for Supermarket from https://github.com/AdExNetwork/aips/issues/31

elpiel:
Closes #340 #312 
For the Input Channel, we have 2 different variants of a Channel:
1. Full Channel
2. Channel returned from Market & Supermarket (with less information that the Full Channel)

Because of this the variables `eventMinPrice` & `eventMaxPrice` only exist if the Full Channel is set to the Input.

### Supermarket PR https://github.com/AdExNetwork/adex-supermarket/pull/29


simzzz (approved):

elpiel:
🙌 thanks @simzzz

adview-manager/src/lib.rs
```diff
@@ -474,15 +468,17 @@ impl Manager {
                 if block_on(self.is_campaign_sticky(campaign.channel.id)) {
                     return vec![];
                 }
-                let mut campaign_input = targeting_input.clone();
```
shirshak55, line 477: Is clone required here?

adview-manager/src/lib.rs
```diff
@@ -474,15 +468,17 @@ impl Manager {
...
-                campaign_input.channel = Some(campaign.channel.clone());
+
+                let campaign_input = targeting_input.clone().with_market_channel(campaign.channel.clone());
 
                 campaign
                     .units_with_price
                     .iter()
                     .filter(|unit_with_price| {
-                        campaign_input.ad_unit = Some(unit_with_price.unit.clone());
+                        let mut unit_input = campaign_input.clone();
```
shirshak55, line 480: is clone required here?

elpiel: It's not required indeed, good catch, thanks!

shirshak55: :)

samparsky (approved):

simzzz (approved):

Issue #330, opened 2020-09-02
(UPDATE to 1.48 when released) Rust 1.46 breaks the build, while 1.45 works just fine


elpiel:
# Build `Error` on Sentry, Rust 1.46:

```
elpiel@kracho:~/PROJECTS/AdEx/rust-adex-validator-stack$ cargo build -p sentry
   Compiling primitives v0.1.0 (/home/elpiel/PROJECTS/AdEx/rust-adex-validator-stack/primitives)
   Compiling adapter v0.1.0 (/home/elpiel/PROJECTS/AdEx/rust-adex-validator-stack/adapter)
   Compiling sentry v0.1.0 (/home/elpiel/PROJECTS/AdEx/rust-adex-validator-stack/sentry)
error: reached the type-length limit while instantiating `std::future::from_generator::<[s...pter::DummyAdapter>]>>, ()}]>}]>`
  --> /home/elpiel/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/src/libcore/future/mod.rs:59:1
   |
59 | / pub const fn from_generator<T>(gen: T) -> impl Future<Output = T::Return>
60 | | where
61 | |     T: Generator<ResumeTy, Yield = ()>,
62 | | {
...  |
85 | |     GenFuture(gen)
86 | | }
   | |_^
   |
   = note: consider adding a `#![type_length_limit="1269353"]` attribute to your crate

error: aborting due to previous error

error: could not compile `sentry`.

To learn more, run the command again with --verbose.
```

## :heavy_check_mark:  Revert it Rust 1.45
For now the solution is to revert Rust version to `1.45.2` and upgrade later

- PR #331 - `rust-toolchain` locked to 1.45.2

## TODO:  Update to 1.48 when released and see this issue is fixed

Related merged PR: rust-lang/rust#78410

### Related Rust bugs:

- https://github.com/rust-lang/rust/issues/75992
- https://github.com/rust-lang/rust/issues/72408

elpiel:
rust-lang/rust#75992 was resolved by reverting some commits in https://github.com/rust-lang/rust/pull/78410

For now we are waiting for the next Rust release, probably 1.48 that will be released this(?) week

shirshak55:
@elpiel its released yesterday so should be fixed.

elpiel:
Thank you for letting me know! I'm already working on updating to it

shirshak55:
[...]

PR #348, opened 2020-11-17
Fix ethereum adapter & update Dockerfile Rust version to 1.45.2

elpiel:


Dockerfile
```diff
@@ -1,6 +1,6 @@
-FROM rust:1.40 as builder
+FROM rust:1.45.2 as builder
```
shirshak55, line 1: why no use latest version released yesterday? Or is it substrate issue afaik they has some issue which required 2020-10-05 version. But I think I would wait for latest version of compiler ?

elpiel: I was going to update it, however clippy is panicing, see my comment here: https://github.com/AdExNetwork/adex-validator-stack-rust/issues/330#issuecomment-730882481

We don't use substrate in this repo, the issue arouse from compiler regression, this is why we reverted to 1.45.2 in the first place, I just forgot to update the `Dockerfile`

elpiel:
Upgrading to 1.48.0 + some improvements to make `clippy` not `panic!`ing are implemented in #349

Dockerfile
elpiel: See #349 for the `1.48.0` upgrade

PR #352, opened 2020-12-07
Issue 335 update targeting event

simzzz:


sentry/src/event_aggregator.rs
```diff
@@ -143,6 +143,12 @@ impl EventAggregator {
             _ => ResponseError::BadRequest(e.to_string()),
```
elpiel, line 143: Can you also handle the error `OnlyCreatorCanUpdateTargetingRules`

sentry/src/event_aggregator.rs
```diff
@@ -143,6 +143,12 @@ impl EventAggregator {
...
         })?;
 
+        events.iter()
+            .for_each(async |e| match e {
+                Event::UpdateTargeting { targeting_rules } => update_targeting_rules(&app.pool, &channel_id, &targeting_rules).await?,
+                _ => false,
```
elpiel, line 149: Should we use the `bool`? I think we should handle if it was successfully updated in the DB or not

elpiel (commented):
There was an Event that was no longer used iirc
can you double check if it's true and remove it?

sentry/src/event_aggregator.rs
```diff
@@ -138,11 +139,25 @@ impl EventAggregator {
...
+        let events_stream = stream::iter(events);
+        let events_eval = events_stream.then(|ev| async move {
+            match ev {
+                Event::UpdateTargeting { targeting_rules } => {
+                    update_targeting_rules(&app.pool, &channel_id, &targeting_rules).await
+                }
+                _ => Ok(false),
+            }
+        });
+        events_eval.collect::<Vec<_>>().await;
```
elpiel, line 159: cool! I didn't know about this trick with `Stream`.

does `collect()` creates a `Vec<Result<bool, ...>>` in this case? Maybe we can collect in single `Result` and check if it's `Ok(true)` otherwise we can return a `400 Bad Request` maybe?

simzzz: Yes, it is `Vec<Result<bool, ...>>`, if it is`Ok(false)` it means it's just not an `UpdateTargeting` event.

elpiel (changes requested):
Looks almost ready :+1: 

Regarding testing and Redis, check this issue that can be tackled separately: https://github.com/AdExNetwork/adex-validator-stack-rust/issues/357

elpiel (approved):

PR #369, opened 2021-01-18
Run validator tests

samparsky:
Adds the ability to run JS validator tests against Rust validator.

- Adds JS validator as a Git submodule

### DIFF
--- a/run-tests.sh
+++ b/run-tests.sh
@@ -1,0 +1,32 @@
+#!/bin/sh
+
+function setup() {
+    cd ./lib/validator && npm install
+}
+
+SUBCOMMAND=$1
+
+# set rust validator worker env var
+export RUST_VALIDATOR_WORKER=./target/debug/validator_worker 
+
+# RUN Tests
+if [ "$SUBCOMMAND" == "--setup" ]; then
+    echo "Setup validator tests"
+    setup    
+else
+    cargo build -p validator_worker
+    docker-compose -f docker-compose.dev.yml up
+
+    echo "Running route and integration tests"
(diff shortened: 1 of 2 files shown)


run-tests.sh
```diff
@@ -0,0 +1,32 @@
+#!/bin/sh
+
+function setup() {
+    cd ./lib/validator && npm install
```
elpiel, line 4: this relies on NPM and node locally. please use Docker.

run-tests.sh
```diff
@@ -0,0 +1,32 @@
...
+
+# set rust validator worker env var
+export RUST_VALIDATOR_WORKER=./target/debug/validator_worker 
+
+# RUN Tests
+if [ "$SUBCOMMAND" == "--setup" ]; then
+    echo "Setup validator tests"
+    setup    
+else
+    cargo build -p validator_worker
```
elpiel, line 17: NOTE: this might not play well with CI because of the `target` it's building for.

run-tests.sh
```diff
@@ -0,0 +1,32 @@
...
+    docker-compose -f docker-compose.dev.yml up
+
+    echo "Running route and integration tests"
+    ./lib/validator/test/routes.js #&& ./lib/validator/test/integration.js
```
elpiel, line 21: One more thing I noticed is that the tests depend on configuration values from JS.
Since we are running the Rust validator, they might differ, as they do now for e.g. `wait_time`, `aggr_throttle`

samparsky: those vars shouldn't matter as we are running the tests in development mode

PR #372, opened 2021-01-22
Running sentry tests no longer hits redis rate limit

### LINKED ISSUE #357: RateLimit and redis randomly failing tests
We currently hit the `RateLimit` from `sentry::access::check_access` in tests (ran asynchronously) quite often as it's they are timed tests.

* [ ] Make sure for each test in `access.rs` to setup a different Database in order to not have random failing tests
* [ ] improve the usability of `redis_connection`
* [ ] enable ignored tests from PR #352  (issue #335)

simzzz:
closes #357 
After some research, I decided to go with the following changes:

- Use `SELECT` to switch the Redis DB for each test, making sure tests don't mess it up for one another
- Since tests still hit the rate limit even with `SELECT` I discovered that a common approach is running the tests in the database module sequentially. I used the `serial_test` crate for that.


elpiel (changes requested):
This will still have issues because of the `auth` tests.

Make sure to create a `Pool` struct for the tests that will automatically release the Database once the test is done with them.

it should also Flush the Database before releasing it to the pool

https://docs.rs/deadpool

elpiel (approved):
LGTM! Should be enough for our needs atm

PR #376, opened 2021-02-19
Issue 319 event aggregator changes 

simzzz:


sentry/src/event_reducer.rs
```diff
@@ -1,18 +1,28 @@
...
     channel: &Channel,
     initial_aggr: &mut EventAggregate,
     ev: &Event,
+    payout: &PayoutResult,
     session: &Session,
 ) -> Result<(), Box<dyn std::error::Error>> {
     let event_type = ev.to_string();
+    if let Ok(Some(payout)) = payout {
+        info!(&logger, "Event type {}, with payout Validator ID: {}, and Pricing of {}", event_type, payout.0, &payout.1.to_string());
+        let cloned_payout = (payout.0, payout.1.clone());
```
simzzz: There's probably a better way to do this. Maybe `PayoutResult` should become its own struct?

sentry/src/event_aggregator.rs
```diff
@@ -155,19 +155,32 @@ impl EventAggregator {
...
+                &payouts[i],
                 &session,
             ) {
                 Ok(_) => {}
                 Err(err) => error!(&app.logger, "Event Reducer failed"; "error" => ?err ),
             }
         });
 
+        // We don't want to save empty aggregates
+        if event_reducer::is_empty(&record.aggregate) {
```
elpiel: You don't need a separate function for this, the hasmap already includes one:
https://doc.rust-lang.org/std/collections/struct.HashMap.html#method.is_empty

sentry/src/event_aggregator.rs
```diff
@@ -155,19 +155,32 @@ impl EventAggregator {
             update_targeting_rules(&app.pool, &channel_id, &new_rules).await?;
         }
 
-        events.iter().for_each(|ev| {
+        // Pre-computing all payouts once
+        let payouts: Vec<PayoutResult> = events
+            .iter()
+            .filter(|ev| ev.is_click_event() || ev.is_impression_event())
```
elpiel, line 161: Let's break down what's happening here...
First we filter in only the events that are click or impression and we calculate the payout. This however will result in the `collect()` creating a new Vec with totally different indices, that later **won't** match the `events.iter()` indices.

So this is a "Big no-no" :sweat_smile: .
It's also good to avoid `somethin[i]` in these situations, because we can't be certain that all the indices will align and will be there, which will cause the program to crash ( trying to access it with `[index]`).


What you could do instead, which is a bit different from the JS impl, is to bind them together with for example a tuple `(event, payout)`. In this case, the payout could be an `Option`. This way we can distinguish between the Payout events (click and impression) and skip those that aren't (that are `None`).

sentry/src/payout.rs
```diff
@@ -9,9 +9,14 @@ use primitives::{
 use slog::{error, Logger};
 use std::cmp::{max, min};
 
-type Result = std::result::Result<Option<(ValidatorId, BigNum)>, Error>;
-
-pub fn get_payout(logger: &Logger, channel: &Channel, event: &Event, session: &Session) -> Result {
+pub type PayoutResult = std::result::Result<Option<(ValidatorId, BigNum)>, Error>;
```
elpiel: There is actually no need to rename the result. It's a common practice to have a `type Result` which is used for a whole module and it's easy to rename it when importing it `use payout::Result as PayoutResult` or just use the module as prefix `result_var: payout::Result`

You can check out this guideline, which I don't know by heart, but it's very nice reference for e.g. naming functions, methods, etc.
https://rust-lang.github.io/api-guidelines/

sentry/src/event_reducer.rs
elpiel: I am not really certain why do we do have to duplicate logic. The two matches bellow already have the same logic implemented without using `[event_type]` but `initial_aggr.events.get(&event_type)`.
You need to link the payout from the function to the event. See my other comment for using a tuple for example.

sentry/src/event_aggregator.rs
```diff
@@ -155,19 +155,32 @@ impl EventAggregator {
...
+            .map(|ev| get_payout(&app.logger, &record.channel, ev, &session))
+            .collect();
```
elpiel: Keep in mind, that we use `let payout = get_payout(logger, &channel, &ev, session)?;` in the `fn reduce()` which means that if one of the `get_payout` calls fails, then the whole `reduce()` fn. will fail (return an error). This should also be the case when we calculate the payouts here.

This is easily mitigated by: https://hoverbear.org/blog/a-useful-error-pattern/
If one of the `Result`s in the iterator is `Err` then the whole `collect` will return an `Err`. It will even error on the first `Err` that it encounters (called "short circuiting"), which means that it won't waste time to calculate all results first and then return an error.

elpiel (changes requested):

sentry/src/event_aggregator.rs
```diff
@@ -155,19 +155,35 @@ impl EventAggregator {
             update_targeting_rules(&app.pool, &channel_id, &new_rules).await?;
         }
 
-        events.iter().for_each(|ev| {
+        // Pre-computing all payouts once
+        let payouts: Vec<(&Event, Option<(ValidatorId, BigNum)>)> = events
```
elpiel: instead of iterating twice, use the map to both calculate the payout and handle the event

sentry/src/event_reducer.rs
```diff
@@ -104,15 +102,8 @@ mod test {
...
-        let session = Session {
-            ip: Default::default(),
-            country: None,
-            referrer_header: None,
-            os: None,
-        };
-
         for i in 0..101 {
-            reduce(&logger, &channel, &mut event_aggr, &event, &session)
+            reduce(&channel, &mut event_aggr, &event, &None)
```
elpiel: Uhhh... ohhh... this should be fixed. The payout here should be 1.
Bellow where the `even_payouts` is fetched and `assert!`ed it actually gets the `impression_event.event_counts` as well, which is a typo!

elpiel (changes requested):

sentry/src/event_reducer.rs
```diff
@@ -101,10 +101,10 @@ mod test {
             ad_slot: None,
             referrer: None,
         };
-
+        let payout = Some((IDS["publisher"], BigNum::from(1)));
         for i in 0..101 {
-            reduce(&channel, &mut event_aggr, &event, &None)
-                .expect(&format!("Should be able to reduce event #{}", i));
```
elpiel, line 116: this is actually the same. Expect will unwrap or it will panic, no need to change this.

elpiel (changes requested):

elpiel (approved):

Issue #382, opened 2021-03-09
AIP #61 v5: Channel and Campaign


elpiel:
## [AIP AmbireTech/adex-validator-stack-rust#61](https://github.com/AdExNetwork/aips/issues/61)

Part of AmbireTech/adex-validator-stack-rust#377 

### Channel

As AIP#61 states the Channel is defined by `(leader, follower, guardian, tokenAddr, nonce)`

- Channel validation - is token allowed? is the requester a Leader or a Follower?
- No routes for storing Channels, they are gone.
- Only routes for listing all the channels (fetched by unique `channel_id` in the `campaigns` table)

### Campaign
- A Campaign is defined by `Campaign { id: CampaignId, channel: Channel, creator, validators, title.... rest of the ChannelSpec fields }`
  Note: The fields were aligned with Ivo before making the changes for the Campaign.
- `CampaignId` should be a randomly assigned value by `Sentry`, we've chosen the `Uuid` as a random generator in Rust and uses a representation of the 16 bytes as `0x` prefixed hex string

#### No migration

Apart from that, we can remove obsolete fields in the `ChannelSpec` (v4) as the v5 will not have migration from v4 as it will work in a different way than v4.

### Tasks:
- [x] v5 Campaign & Channel PRs AmbireTech/adex-validator-stack-rust#383, AmbireTech/adex-validator-stack-rust#388 (commit https://github.com/AdExNetwork/adex-validator-stack-rust/pull/388/commits/55616d4d6e4304d7c2672a1a25cd3d3206c07b23)
- [x] Campaign Validator AmbireTech/adex-validator-stack-rust#400 
- [x] Minimum token deposit & minimum validator fee AmbireTech/adex-validator-stack-rust#400 

**Sentry**
- [x] Database changes (migrations, queries, etc)
  - [x] Campaign DB table - PR AmbireTech/adex-validator-stack-rust#388 commit https://github.com/AdExNetwork/adex-validator-stack-rust/pull/388/commits/55616d4d6e4304d7c2672a1a25cd3d3206c07b23
  - [x] Other DB tables PR AmbireTech/adex-validator-stack-rust#388 commit https://github.com/AdExNetwork/adex-validator-stack-rust/pull/388/commits/55616d4d6e4304d7c2672a1a25cd3d3206c07b23
  - [x] Insert/Fetch Campaign AmbireTech/adex-validator-stack-rust#400 
[...]

PR #383, opened 2021-03-15
AdEx v5: Issue #382 Channel and Campaign

elpiel:
This PR tackles some of the changes required for #382.

#### Changelog:

Introduces the new structs & enums for AdEx v5:
- `primitives::campaign::{Campaign, Validators, Active, etc.}`
- `primitives::channel_v5::Channel` (should replace the `channel` module which has the v4 `Channel`)
  - implemented `fn id(&self) -> ChannelId`
  - `primitives::channel_v5::Nonce` - a `U256` with a string (de)serializatoin instead of hex
- `Address` - should replace all the instances of `ValidatorId` where the `ValidatorId` is used for an Address (token address, contract address, spender, earner, etc.)
- `primitives::ToHex` - trait for encoding values to `0x` prefixed or non-prefixed `Strings`
- `primitives::eth_checksum` - now uses `tiny_keccak`
- Cargo - `primitives`:
  - Updated `tiny_keccak` to `2.0`
  - Removed `rust-crypto` dependency
  - Updated `num` to `0.4` and use the feature `num-bigint`
  - Removed `num-bigint` dependency
 - Cargo - `validator_worker`:
   - Updated `num` to `0.4`

### DIFF
--- a/primitives/src/campaign.rs
+++ b/primitives/src/campaign.rs
@@ -60,7 +60,7 @@ pub struct Active {
 impl Campaign {
     /// Matches the Channel.leader to the Campaign.spec.leader
     /// If they match it returns `Some`, otherwise, it returns `None`
-    pub fn leader<'a>(&'a self) -> Option<&'a ValidatorDesc> {
+    pub fn leader(&self) -> Option<&'_ ValidatorDesc> {
         if self.channel.leader == self.validators.leader().id {
             Some(self.validators.leader())
         } else {
@@ -70,7 +70,7 @@ impl Campaign {
 
     /// Matches the Channel.follower to the Campaign.spec.follower
     /// If they match it returns `Some`, otherwise, it returns `None`
-    pub fn follower<'a>(&'a self) -> Option<&'a ValidatorDesc> {
+    pub fn follower(&self) -> Option<&'_ ValidatorDesc> {
         if self.channel.follower == self.validators.follower().id {
             Some(self.validators.follower())
         } else {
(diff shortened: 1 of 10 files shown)


primitives/src/channel_v5.rs
```diff
@@ -0,0 +1,15 @@
+use serde::{Deserialize, Serialize};
+
+use crate::{BigNum, ChannelId, ValidatorId, Address};
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq)]
+#[serde(rename_all = "camelCase")]
+pub struct Channel {
+    pub id: ChannelId,
+    pub leader: ValidatorId,
+    pub follower: ValidatorId,
```
samparsky, line 11: Currently `validatorId === Address`, so why not make `ValidatorId` or `Address`  a type alias?

primitives/src/campaign.rs
```diff
@@ -0,0 +1,239 @@
...
+            None
+        }
+    }
+}
+
+#[derive(Debug, Serialize, Deserialize)]
+pub struct CampaignSpec {
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub title: Option<String>,
+    pub validators: Validators,
```
samparsky: Do we need to specify validators, we already have the Channel?

samparsky: @Ivshti I think it's safe to remove the validators array from the campaign spec. Maybe add `channel id` ?

samparsky (changes requested):

primitives/src/channel_v5.rs
elpiel: We do not want to be able to substitute a `ValidatorId` with an `Address`, `ValidatorId` != `Address`.
If you need to access the bytes or even convert the ValidatorId to the inner Address, that's fine, but it must be explicit.

primitives/src/campaign.rs
elpiel: The v5 `Channel` only contains the addresses of the validators, the Campaign Validators include the `fee`, `fee_addr`, `fee_token`, `url`

simzzz (approved):

samparsky (approved):

Issue #384, opened 2021-03-15
AIP #61 v5: Unified precision (8)


elpiel:
## https://github.com/AdExNetwork/aips/issues/61

Create a Unified Precision number (`UnifiedNum`) with precision of 8 that will be used for all calculations, except when we are creating the MerkleTree, where we need the actual precision of the token in order of the MerkleTree to be calculated properly.

- [x] (PR #385) Token configuration of actual precision (in the configuration of the whitelisted token addresses) 
- [x] Use Unified Precision - `UnifiedNum` (including DB) & floor all the calculations
- [x] Postgres type: `bigint`
- [x] (PR #412) Targeting rules DSL - use UnifiedNum instead of BigNum (see https://github.com/AdExNetwork/adex-protocol/issues/67)

elpiel:
Postgres: Since we are using a `u64` to store the `UnifiedNum` we can use `bigint` (max of `9_223_372_036_854_775_807`) and although we can handle half of what `u64::MAX` = `184_467_440_737.09551615`, it should not cause any issues.
Otherwise an error will be raised by the database and caught in Sentry.

PR #385, opened 2021-03-16
Adapter calculate and retrieve total spendable amount

### LINKED ISSUE #380: AIP #61 v5: Adapter - total spendable amount

> function to calculate total spendable amount; this will be used by sentry to update it on request 

Function: `adapter.get_deposit(channel, address) -> Deposit {total, stillOnCreate2}`
This will be stored in the DB as ` spender, channel, channel_id, total, stillOnCreate2 } ` later on, unique  by `channel_id, spender` #381 

- [x] outpace.deposited[channelId][spender] ( `outpace.deposited(channelId, spender)` )
- [x] `stillOnCreate2` - `erc20.balanceOf(counterfactualDepositAddr)` this should be passed by the `channel.deposit_asset`
  create2 address: https://github.com/AdExNetwork/adex-protocol-eth/blob/identity-tests/contracts/IdentityFactory.sol#L39-L41
- [x] Configurable minimum deposit (per whitelisted stablecoin)
- [x] `store both values so that we know stillOnCreate2`
- [x] update git submodule to get access to https://github.com/AdExNetwork/adex-protocol-eth/blob/master/abi/ERC20.json

simzzz:
Implemented `adapter.get_deposit(...)`for the Ethereum and Dummy adapters, including an integration test.

## Change Log:

- `adapter::EthereumAdapter` - Added a `get_deposit()` function which calculates and returns total spendable amount, using the `OUTPACE` and `ERC20` ABI's.
- `adapter::ethereum::Error` - Added two new possible errors. The first being `TokenNotWhitelisted` which is returned when the function is called with a channel with a deposit asset not being in the config's whitelist. The second is `InvalidDepositAsset` which is used when a channel's deposit asset cannot be converted to `Address`. This error is here temporarily and I assume it will be gone soon when the `channel_v5` replaces the old `channel` in the adapter.
- `dev.toml and prod.toml configs` - Changed the structure of `token_addr_whitelist` from an array of strings, to an array with objects that contain `address`, `min_token_units_for_deposit` and `precision`.
- `primitives::adapter::Deposit` - new struct representing the `get_deposit()` output. Added signature for `get_deposit` in the `impl`
- `primitives::channel_validator::asset_listed()` - Changed `asset_listed()` function to work with the new token whitelist structure from String Vec to HashMap.
- `primitives::config` - Changed token whitelist from String Vec to HashMap. Added a new struct called `TokenInfo` which represents the value for the token whitelist HashMap.


primitives/src/adapter.rs
```diff
@@ -6,11 +6,17 @@ use serde::{Deserialize, Serialize};
 use std::collections::HashMap;
 use std::convert::From;
 use std::fmt;
+use web3::types::U256;
 
 pub type AdapterResult<T, AE> = Result<T, Error<AE>>;
 
 pub trait AdapterErrorKind: fmt::Debug + fmt::Display {}
 
+pub struct SpendableOutput {
```
elpiel: I think that just `Spendable` is clear enough.

I am not 100% sure if `U256` will remain as type here, we will either use a `BigNum` and include the token info (like the precision) or we will use a `UnifiedNum` (precision 8) #384

primitives/src/channel_validator.rs
```diff
@@ -96,11 +97,11 @@ pub fn creator_listed(channel: &Channel, whitelist: &[ValidatorId]) -> bool {
     whitelist.is_empty() || whitelist.iter().any(|allowed| allowed.eq(&channel.creator))
 }
 
-pub fn asset_listed(channel: &Channel, whitelist: &[String]) -> bool {
+pub fn asset_listed(channel: &Channel, whitelist: &HashMap<String, TokenInfo>) -> bool {
```
elpiel: When #382 is merged we can use the `Address` type instead of `String`, or we can use `ValidatorId` for the time being and later change it once the PR is merged.

primitives/src/config.rs
```diff
@@ -14,6 +15,12 @@ lazy_static! {
             .expect("Failed to parse prod.toml config file");
 }
 
+#[derive(Serialize, Deserialize, Debug, Clone)]
+pub struct TokenInfo {
+    pub min_deposit: u32,
```
elpiel: I am not sure if this will be enough. It matches ideally for DAI (18 decimals) since a u32::MAX is `4_294_967_295` (i.e. minimum deposit that will convert to `42` in 8 decimal precision). Maybe we can use BigNum instead (this means that the config value should be a string)?

adapter/src/ethereum/error.rs
```diff
@@ -22,6 +22,7 @@ pub enum Error {
     ContractQuerying(web3::contract::Error),
     /// Error occurred during verification of Signature and/or StateRoot and/or Address
     VerifyAddress(VerifyError),
+    TokenNotWhitelisted(String),
```
elpiel: This should probably also change to `Address` when we merge the other PR

adapter/src/ethereum.rs
```diff
@@ -269,6 +271,64 @@ impl Adapter for EthereumAdapter {
...
+        let outpace_contract = Contract::from_json(
+            self.web3.eth(),
+            self.config.ethereum_core_address.into(),
+            &OUTPACE_ABI,
+        )
+        .map_err(Error::ContractInitialization)?;
+
+        let erc20_contract = Contract::from_json(
+            self.web3.eth(),
+            self.config.ethereum_core_address.into(),
```
elpiel: this should be the Token address instead of the core one.

adapter/src/ethereum.rs
```diff
@@ -269,6 +271,64 @@ impl Adapter for EthereumAdapter {
...
+            &ERC20_ABI,
+        )
+        .map_err(Error::ContractInitialization)?;
+
+        let mut amount: U256 = tokio_compat_02::FutureExt::compat(async {
+            tokio_compat_02::FutureExt::compat(outpace_contract.query(
+                "deposits",
+                (channel.id.into_token(), spender.to_hex_prefix_string().into_token()),
```
elpiel: You need these types to be from the following list:

https://docs.rs/web3/0.15.0/web3/contract/tokens/trait.Tokenize.html?search=into_token

- The `ChannelId` has `[u8; 32]` (`FixedBytes` or known as `bytes32` in the contact )
- The `ValidatorId` hsa `[u8; 20]` (or `Address`)
You can also see those types in the contract:

https://github.com/AdExNetwork/adex-protocol-eth/blob/identity-tests/contracts/OUTPACE.sol#L43
It's possible that the this will fail if you try to test it.

adapter/src/ethereum.rs
```diff
@@ -269,6 +271,64 @@ impl Adapter for EthereumAdapter {
...
+            ))
+            .await
+        })
+        .await
+        .map_err(Error::ContractQuerying)?;
+
+        let to_be_deposited: U256 = tokio_compat_02::FutureExt::compat(async {
+            tokio_compat_02::FutureExt::compat(erc20_contract.query(
+                "balanceOf",
+                channel.deposit_asset.clone().into_token(),
```
elpiel: This should probably also change, because we use `Address` for the token address in the new `Channel`.
The types right now are different, the contract expects an `Address`, however we pass it a `Token::String`

You can also check a more human-readable form of the ABI here: https://ethereumdev.io/abi-for-erc20-contract-on-ethereum/

elpiel (commented):
Looks like you have great progress on the task already :tada: 
Can you also make sure to include a list of the Changelog in PR so we can later add it to the CHANGELOG.md file.

elpiel:
After some discussion with Ivo, we came to the following changes:
`get_deposit(channel: &Channel, user: &Address) -> Deposit`
`struct Deposit { total: BigNum, pending: BigNum }`

This could change in the future, as we will introduce the `UnifiedNum` and we have yet to see if we will use the full precision from the validator (`BigNum`) or directly work with the `UnifiedNum` from the `Adapter` onward.

adapter/src/ethereum.rs
```diff
@@ -269,6 +273,77 @@ impl Adapter for EthereumAdapter {
         ewt_sign(&wallet, &self.keystore_pwd, &payload)
             .map_err(|err| AdapterError::Adapter(Error::SignMessage(err).into()))
     }
+
+    async fn get_deposit(
+        &self,
+        channel: &Channel,
```
elpiel, line 272: you can now use the new channel_v5::Channel`

primitives/Cargo.toml
```diff
@@ -54,6 +54,7 @@ futures = "0.3"
 async-trait = "0.1"
 # Other
 lazy_static = "1.4.0"
+web3 = { version = "0.15" }
```
elpiel: do we use web3 in primitives? If now we can remove it.

primitives/src/config.rs
```diff
@@ -34,14 +44,44 @@ pub struct Config {
...
+fn deserialize_token_whitelist<'de, D>(
+    deserializer: D,
+) -> Result<HashMap<Address, TokenInfo>, D::Error>
+where
+    D: Deserializer<'de>,
+{
+    let mut tokens_whitelist = HashMap::new();
+    let array: Vec<ConfigWhitelist> = Deserialize::deserialize(deserializer)?;
+
+    array.into_iter().for_each(|i| {
```
elpiel: you can actually `collect()` a tuple of `(config_whitelist.address, token_info)` directly into a `HashMap`. Small trick to remove the need of insertion, new mut hashmap and etc.
You would also need to use `.map()` instead of `for_each`.
Another option is `fold()`, but in this case `.map().collect()` will be least amount code.

primitives/src/config.rs
```diff
@@ -34,14 +44,44 @@ pub struct Config {
...
     pub ethereum_network: String,
     pub ethereum_adapter_relayer: String,
     pub validators_whitelist: Vec<ValidatorId>,
 }
 
+#[derive(Serialize, Deserialize, Debug, Clone)]
+struct ConfigWhitelist {
+    address: Address,
+    min_token_units_for_deposit: BigNum,
+    precision: NonZeroU8,
```
elpiel, line 64: could you make `precision` & `decimals` have the same name? this will help out with confusion.
Also `min token units for deposit` might be a good idea to be in `UnifiedNum` since the only place where we would use BigNum should be in the merkle tree & adapter

elpiel (commented):
Can you also make sure to fix the gitmodule

adapter/src/ethereum.rs
```diff
@@ -269,6 +278,71 @@ impl Adapter for EthereumAdapter {
...
+
+        let erc20_contract =
+            Contract::from_json(self.web3.eth(), channel.token.as_bytes().into(), &ERC20_ABI)
+                .map_err(Error::ContractInitialization)?;
+
+        let total: H256 = tokio_compat_02::FutureExt::compat(async {
+            tokio_compat_02::FutureExt::compat(outpace_contract.query(
+                "deposits",
+                (
+                    H256(*channel.id()).into_token(),
```
elpiel: this is not really readable, could you used the `Token::Address`, for the ChannelId we could leave it as it is. It wont really give us any benefit to use `Token::FixedBytes` (which is the equivalent of bytes32, bytes64, etc. in solidity)

adapter/src/ethereum.rs
```diff
@@ -269,6 +278,71 @@ impl Adapter for EthereumAdapter {
...
+                None,
+            ))
+            .await
+        })
+        .await
+        .map_err(Error::ContractQuerying)?;
+
+        let mut total = BigNum::from_str(&total.to_string())?;
+
+        let still_on_create_2: H256 = tokio_compat_02::FutureExt::compat(async {
```
elpiel: Should be https://docs.rs/ethabi/13.0.0/ethabi/type.Uint.html

adapter/src/ethereum.rs
```diff
@@ -269,6 +278,71 @@ impl Adapter for EthereumAdapter {
...
+            tokio_compat_02::FutureExt::compat(erc20_contract.query(
+                "balanceOf",
+                Bytes(channel.token.as_bytes().to_vec()).into_token(),
```
elpiel: `EthAddress` (you can use the `channel.token.to_bytes()` which gives you `[u8; 20]`)

adapter/src/ethereum.rs
```diff
@@ -269,6 +278,71 @@ impl Adapter for EthereumAdapter {
...
+                None,
+                Options::default(),
+                None,
+            ))
+            .await
+        })
+        .await
+        .map_err(Error::ContractQuerying)?;
+
+        let still_on_create_2 = BigNum::from_str(&still_on_create_2.to_string())?;
```
elpiel: you can use `still_on_create_2.to_string().parse()` this is what `FromStr` allows you to do. You can use it directly like this, but it's meant more like a syntax sugar for String & &str `.parse()`

adapter/src/ethereum/error.rs
```diff
@@ -1,5 +1,6 @@
 use primitives::adapter::{AdapterErrorKind, Error as AdapterError};
-use primitives::ChannelId;
+use primitives::address::Error as AddressError;
```
elpiel: can you group them together from now on?
E.g. `primitives::{ all imports here }`;

docs/config/dev.toml
```diff
@@ -26,5 +26,5 @@ ethereum_adapter_relayer = 'https://goerli-relayer.adex.network'
 creators_whitelist = []
 minimal_deposit = "0"
 minimal_fee = "0"
-token_address_whitelist = []
 validators_whitelist = []
+token_address_whitelist = []
```
elpiel: please add tokens to the whitelist for the known used stablecoins we use: DAI, USDT, USDC.
Also if this is a Table could you please use the alternative styling of:
```
[[token_address_whitelist]]
address = ...
min_token_units_for_deposit = ...
...
[[token_address_whitelist]]
address = ...
min_token_units_for_deposit = ...
...

```

docs/config/prod.toml
```diff
@@ -27,3 +27,4 @@ minimal_deposit = "0"
 minimal_fee = "0"
 token_address_whitelist = ['0x89d24A6b4CcB1B6fAA2625fE562bDD9a23260359', '0x6B175474E89094C44Da98b954EedeAC495271d0F']
```
elpiel: shouldn't this be dumped now?

adapter/src/dummy.rs
```diff
@@ -129,4 +130,16 @@ impl Adapter for DummyAdapter {
             ))),
         }
     }
+
+    async fn get_deposit(
+        &self,
+        channel: &ChannelV5,
+        address: &Address,
+    ) -> AdapterResult<Deposit, Self::AdapterError> {
+        // TODO - complete function
```
elpiel: any ideas on how we should do this in the DummyAdapter?

elpiel (changes requested):
Also I don't see any tests for the new method of the ethereum adapter

adapter/src/ethereum.rs
```diff
@@ -269,6 +296,69 @@ impl Adapter for EthereumAdapter {
...
+                None,
+            ))
+            .await
+        })
+        .await
+        .map_err(Error::ContractQuerying)?;
+
+        let mut total = BigNum::from_str(&total.to_string())?;
+
+        let counterfactual_address = get_counterfactual_address(channel.token);
```
samparsky: The counterfactual address doesn't depend on the `channel.token`, this should be `address` param

adapter/src/ethereum.rs
```diff
@@ -82,6 +91,24 @@ impl EthereumAdapter {
...
+    init_code.update(b"0x00");
+    let mut hashed_init_code: [u8; 32] = [0; 32];
+    init_code.finalize(&mut hashed_init_code);
+
+    let mut result = Keccak::new_keccak256();
+    result.update(b"0xff");
+    result.update(token.as_bytes());
+    result.update(&salt);
+    result.update(&hashed_init_code);
+    let mut res: [u8; 20] = [0; 20];
```
samparsky: we need from the `12th byte downward` and not the `first 20 bytes`. You can checkout the specification here https://eips.ethereum.org/EIPS/eip-1014

adapter/src/ethereum.rs
```diff
@@ -82,6 +91,24 @@ impl EthereumAdapter {
     }
 }
 
+fn get_counterfactual_address(token: Address) -> H160 {
```
samparsky: This requires some changes

This should accept the`channel struct`, `outpace address` and `depositor`.
We then `abi.encode` the params and append it to the bytecode of the depositor contract here https://github.com/AdExNetwork/adex-protocol-eth/blob/master/contracts/deposits/Depositor.sol that's what forms the `initcode` 

You can check out this article for more clarification https://medium.com/@martinetlee/notes-on-eip-1014-create2-1-351c47e22a89

samparsky: You can also checkout this library https://docs.rs/create2/0.0.2/create2/fn.calc_addr_with_hash.html the address param there is the `sweeper` address

samparsky (changes requested):

adapter/src/ethereum.rs
samparsky: To properly test we would have to deploy the sweeper contract too

elpiel:
## TODO:
- [x] `get_deposit` of Dummy adapter

docs/config/dev.toml
```diff
@@ -20,11 +20,30 @@ ip_rate_limit = { type = 'ip', timeframe = 20000 }
 sid_rate_limit = { type = 'sid', timeframe = 20000 }
 
 ethereum_core_address = '0x333420fc6a897356e69b62417cd17ff012177d2b'
+# TODO: Replace with real contract address
+outpace_address = '0x333420fc6a897356e69b62417cd17ff012177d2b'
+# TODO: Replace with real contract address
```
samparsky, line 25: to make tests reproducible we can create a ganache snapshot of the required deployments and use it on startup

samparsky (approved):

docs/config/dev.toml
elpiel: This is a cool idea for enhancement!
How can we do that?

elpiel (approved):

PR #386, opened 2021-03-17
AdEx v5: Issue #384 Unified precision - primitives::UnifiedNum

elpiel:
Changelog:
- Add `primitives::UnifiedNum` for #384 with precision of 8

### DIFF
--- a/primitives/src/unified_num.rs
+++ b/primitives/src/unified_num.rs
@@ -10,7 +10,10 @@ use std::{
 
 use crate::BigNum;
 
-/// Unified precision Number with precision 8
+/// Unified Number with a precision of 8 digits after the decimal point
+/// The number can be a maximum of `u64::MAX` (the underlying type),
+/// or in a `UnifiedNum` value `184_467_440_737.09551615`
+/// The actual number is handled as a unsigned number and only the display shows the decimal point
 #[derive(
     Clone,
     Copy,
--- a/primitives/src/unified_num.rs
+++ b/primitives/src/unified_num.rs
@@ -1,5 +1,6 @@
-use num::{pow::Pow, CheckedSub, Integer, One};
-use num_derive::{Num, NumOps, Zero};
+use num::{pow::Pow, CheckedAdd, CheckedDiv, CheckedMul, CheckedSub, Integer, One};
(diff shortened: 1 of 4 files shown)


samparsky:
what's the reason for having a `Unified precision`?

elpiel:
### Reason
> what's the reason for having a `Unified precision`?

The way I understand it is that this will help us to have a unified representation of values that is good enough for all applications. We won't need more precision than this, it will ease the calculations and we can convert to the on-chain values when we create the MerkleTree we can use the full precision of the stablecoin.

Maybe @Ivshti could give a fuller answer here?

### Why 8 in particular

From AIP#61:
> validator: unified precision: use a unified number of decimals for all numbers (8), and only convert to the proper stablecoin decimals when generating the balance tree; this will ensure that we can handle precision even with USDT (6 decimals); normal int numbers can be used in DBs (easier, can aggregate stuff values faster)


If the CPM price is 0.01 then with a precision of 8 and a fee of 1% we will get a value of 0.00000001 in Unified Precision

Ivshti:
The main reason behind using unified precision is being able to use easy-to-work with and fast types such as u64. 

Furthermore, it keeps numbers in the database consistent if we switch between stablecoins with different decimal precision, and we don't need to account for that when calculating stats/aggregate data.

But most importantly, using unified precision allows us to use native number types in databases (mongo, postres), allowing for much easier updates and aggregations, opening the door towards more analytics features.

There's no real reason for supporting more than 8 decimals.

**EDIT:** sorry, closed by accident

primitives/src/unified_num.rs
```diff
@@ -0,0 +1,317 @@
...
+        // 321.00000777 - should floor to 321.000007 (precision 6)
+        let usdt_unified = UnifiedNum::from(32_100_000_777_u64);
+        let usdt_expected = BigNum::from(321_000_007_u64);
+        assert_eq!(
+            usdt_expected,
+            usdt_unified.to_precision(usdt_precision),
+            "It should floor the result of USDT"
+        );
+
+        // 321.00000999
```
samparsky: should be 999 or 7777?

samparsky (approved):

primitives/src/unified_num.rs
elpiel: good catch!

Issue #381, opened 2021-03-09
AIP #61 v5: Spenders & Address accounting (postgres)


elpiel:
https://github.com/AdExNetwork/aips/issues/61

_NOTE:_ In the database the Accounting is flatten and it was introduced in #413 . This is why we have 2 Accounting structs - one for sentry and one for the database.
When building the sentry `Accounting` as a response we should collect all `Channel` Accounting rows from the Database and produce the balances map (earners & spenders).

### TODOs:

- [x] relate the Accounting with a `spender` (advertiser) & `ChannelId`. It should include `earners` & `spenders` balances that should match, i.e. `sum(earners) == sum(spenders)` PR: #413 
- [x] Accounting PRs: #410 #413 
  - [x] it's own DB collection, includes channel_id and is flat structure (1 accounting = 1 earner or spender for channel)
- [x] Spendable amounts - new collection which is unique by `(channel_id, spender)` and we store the `{ total, stillOnCreate2 }` from the #380 , the collection should store `{ spender, channel, channel_id, total, stillOnCreate2 }`

PR #404, opened 2021-05-25
Issue #381 Accounting aggregation

elpiel:
Part of #381


elpiel:
We have decided to o another route and instead of having aggregation (in memory) we will be using atomic operations (redis, postgres) to update the accounting

PR #408, opened 2021-05-31
V5: Authenticated POST route to modify/submit campaigns

### LINKED ISSUE #409: AIP #61 v5: Sentry POST route for inserting/modifying campaign
From AIP61 Main issue:
> validator: sentry: will have an authenticated POST to modify/submit campaigns
> - check if `campaign.budget >= campaignSpent`; update a key in Redis for the remaining spendable amount
> - get the last approved NewState from the DB and set `totalSpent = lastApprovedNewState.spenders[user]`
> - set `totalDeposited` to total from the Spendable DB entry for this `(channelId, spender)`
> - make sure `sum(campaigns.map(getRemaining)) <= totalDeposited - totalSpent` when updating campaigns; by doing that we ensure that advertisers can't spend more than they have
> - please note that we can't assume `totalSpent` equals the sum of all campaign spendings, because a spending can occur outside of campaign (cases like v4 CLOSE or PAY); so we must always compare remaining funds based on `totalDeposited - lastApprovedNewState.spenders[user]`
> - to close a campaign, just set it's budget to what it's spent so far (so that `remaining == 0`)

**### To-Dos:**
**Logic:**
- [x] Retrieve `campaignSpent` from Redis
- [x] Check if we have reached the budget
- [x] Calculate the remaining spendable amount for the campaign and update it on Redis
- [x] Retrieve the latest new state
- [x] Retrieve the latest Spendable entry from the DB
- [x] Calculate the total remaining  amount for the channel
- [x] ~??? Does the totalRemaining need to be stored in Redis?~ no
- [x] Compare the remaining amount for all campaigns on this channel with totalDeposited - totalSpent and update the campaign on Postgres

**API:**
- [x] Create an authenticated POST route for modifying campaigns `/campaign/{id}/`
    - [x] Create a middleware for checking if a campaign with this ID exists in Postgres
    - [x] Track changes in budget, possibly with a struct that includes all of the editable fields.
- [x] Create an authenticated POST route for creating campaigns `/campaign/
    - [x] Deserialize CreateCampaign struct from the request
    - [x] Check if there is enough remaining deposit (through spendable) for the campaign budget
    - [x] Return a campaign with a randomly generated ID and insert it to Postgres

**Testing:**
- [x] Refactor business logic to make it easily testable
- [x] Additional test cases for creating & modifying campaign
- [x] Test Postgres update functions 
- [x] Test `sentry::routes::campaign::update_campaign::get_delta_budget`

**Additional Tasks:**
- [x] Implement pagination for `get_campaigns_by_channel(...)` 
- [x] **Very important** - when modifying a Campaign make sure to Authenticate the caller AND validate that he can perform the modification (`AuthRequired` & check if Auth.uid = campaign.creator - could be middleware as well)
- [x] Update Spendable entry as stated in AIP 61 `we'll also update the document from the adapter before a user tries to mutate (add/update) campaigns`
  - [x] Campaign creation was updated in #432
  - [x] Campaign modification #434

simzzz:


sentry/src/lib.rs
```diff
@@ -150,6 +152,16 @@ impl<A: Adapter + 'static> Application<A> {
 
                 publisher_analytics(req, &self).await
             }
+            ("/campaign.create", &Method::POST) => {
```
elpiel: POST "/campaign" should be enough.
We can follow the current REST API route scheme.

sentry/src/routes/campaign.rs
```diff
@@ -20,25 +38,110 @@ pub async fn create_campaign<A: Adapter>(
...
-    //             _ => Err(error_response),
-    //         }
-    //     }
-    //     Ok(false) => Err(error_response),
-    //     _ => Ok(()),
-    // }?;
+
+    match insert_or_modify_campaign(&app.pool, &campaign, &app.redis).await {
+        Err(error) => {
+            // error!(&app.logger, "{}", &error; "module" => "create_channel");
```
elpiel: Can you fix the logging? It is still be valuable to log this information.

sentry/src/routes/campaign.rs
```diff
@@ -20,25 +38,110 @@ pub async fn create_campaign<A: Adapter>(
...
+    let key = format!("adexCampaign:campaignSpent:{}", id);
+    // campaignSpent tracks the portion of the budget which has already been spent
+    let campaign_spent = match redis::cmd("GET")
+    .arg(&key)
+    .query_async::<_, Option<String>>(&mut redis.clone())
+    .await?
+    {
+        Some(spent) => UnifiedNum::from(spent),
+        // TODO: Double check if this is true
+        // If the campaign is just being inserted, there would be no entry therefore no funds would be spent
```
elpiel: This might be the place where we load the information from the db and add this value if it does not exist in Redis.

sentry/src/routes/campaign.rs
```diff
@@ -20,25 +38,110 @@ pub async fn create_campaign<A: Adapter>(
...
+    let key = format!("adexCampaign:remainingSpendable:{}", id);
+    redis::cmd("SET")
+        .arg(&key)
+        .arg(amount.to_u64())
+        .query_async(&mut redis.clone())
+        .await?;
+    Ok(true)
+}
+
+async fn update_remaining_for_channel(redis: &MultiplexedConnection, id: ChannelId, amount: UnifiedNum) -> Result<bool, PoolError> {
```
elpiel: Do we need to store this information for the channel? what is it's usecase?

sentry/src/routes/campaign.rs
```diff
@@ -20,25 +38,110 @@ pub async fn create_campaign<A: Adapter>(
...
-    //         }
-    //     }
-    //     Ok(false) => Err(error_response),
-    //     _ => Ok(()),
-    // }?;
+
+    match insert_or_modify_campaign(&app.pool, &campaign, &app.redis).await {
+        Err(error) => {
+            // error!(&app.logger, "{}", &error; "module" => "create_channel");
+            return Err(ResponseError::Conflict("channel already exists".to_string()));
```
elpiel: Channel?

sentry/src/routes/campaign.rs
```diff
@@ -20,25 +38,110 @@ pub async fn create_campaign<A: Adapter>(
...
+        .map(async |c| {
+            let spent = get_spent_for_campaign(&redis, c.id).await?;
+            let remaining = c.budget - spent;
+            remaining
+        })
+        .sum();
+    Ok(sum_of_campaigns_remaining)
+}
+
+pub async fn insert_or_modify_campaign(pool: &DbPool, campaign: &Campaign, redis: &MultiplexedConnection) -> Result<bool, ResponseError> {
```
elpiel: This is not insert or modify. When creating Campaign it generates the ID of the campaign and should return it as a response.
The CreateCamapign doesn't require you to set the id, it's randomly generated on campaign creation.

sentry/src/routes/campaign.rs
```diff
@@ -20,25 +38,110 @@ pub async fn create_campaign<A: Adapter>(
...
+    let campaign_spent = get_spent_for_campaign(&redis, campaign.id).await?;
+
+    // Check if we haven't exceeded the budget yet
+    if campaign.budget <= campaign_spent {
```
elpiel: This should be removed. When creating campaign you should check if there's enough Spendable left for this Channel (`campaign.channel.id()`) and to see if the budget fits in the spendable!

sentry/src/routes/campaign.rs
```diff
@@ -20,25 +38,110 @@ pub async fn create_campaign<A: Adapter>(
...
+
+
+    // Getting the latest new state from Postgres
+    let latest_new_state = latest_new_state_v5(&pool, &campaign.channel, "").await?;
+    // Gets the latest Spendable for this (spender, channelId) pair
+    let latest_spendable = fetch_spendable(pool.clone(), &campaign.creator, &campaign.channel.id()).await?;
+
+    let total_deposited = latest_spendable.deposit.total;
+    let total_spent = if let Some(lns) = latest_new_state {
+        lns.msg.into_inner().spenders[campaign.creator]
```
elpiel: this could panic! Don't use indices if the record can be missing there are methods that return Option instead.

sentry/src/routes/campaign.rs
```diff
@@ -20,25 +38,110 @@ pub async fn create_campaign<A: Adapter>(
...
+
+pub async fn insert_or_modify_campaign(pool: &DbPool, campaign: &Campaign, redis: &MultiplexedConnection) -> Result<bool, ResponseError> {
+    let campaign_spent = get_spent_for_campaign(&redis, campaign.id).await?;
+
+    // Check if we haven't exceeded the budget yet
+    if campaign.budget <= campaign_spent {
+        return Err(ResponseError::FailedValidation("No more budget available for spending".into()));
+    }
+
+    let remaining_spendable_campaign = campaign.budget - campaign_spent;
```
elpiel: this should be `checked_sub()`

sentry/src/routes/campaign.rs
```diff
@@ -20,25 +38,110 @@ pub async fn create_campaign<A: Adapter>(
...
+    let latest_spendable = fetch_spendable(pool.clone(), &campaign.creator, &campaign.channel.id()).await?;
+
+    let total_deposited = latest_spendable.deposit.total;
+    let total_spent = if let Some(lns) = latest_new_state {
+        lns.msg.into_inner().spenders[campaign.creator]
+    } else {
+        0
+    };
+
+    let total_remaining = total_deposited - total_spent;
```
elpiel: use `checked_sub`

sentry/src/routes/campaign.rs
```diff
@@ -20,25 +38,110 @@ pub async fn create_campaign<A: Adapter>(
...
+    Ok(true)
+}
+
+async fn get_campaigns_remaining_sum(redis: &MultiplexedConnection, pool: &DbPool, campaign: &Campaign) -> Result<UnifiedNum, PoolError> {
+    let campaigns_for_channel = get_campaigns_for_channel(&pool, &campaign).await?;
+    let sum_of_campaigns_remaining = campaigns_for_channel
+        .into_iter()
+        .map(async |c| {
+            let spent = get_spent_for_campaign(&redis, c.id).await?;
+            let remaining = c.budget - spent;
```
elpiel: always use `.checked_sub()`

sentry/src/routes/campaign.rs
```diff
@@ -20,25 +38,110 @@ pub async fn create_campaign<A: Adapter>(
...
+            remaining
+        })
+        .sum();
```
elpiel: This `sum()` should return `Option<UnifiedNum>` as it might overflow.
We need to make it a `Result` and handle it

sentry/src/db/campaign.rs
```diff
@@ -49,6 +49,51 @@ pub async fn fetch_campaign(pool: DbPool, campaign: &Campaign) -> Result<Campaig
     Ok(Campaign::from(&row))
 }
 
+pub async fn get_campaigns_for_channel(pool: &DbPool, campaign: &Campaign) -> Result<Vec<Campaign>, PoolError> {
+    let client = pool.get().await?;
+    let statement = client.prepare("SELECT id, channel, creator, budget, validators, title, pricing_bounds, event_submission, ad_units, targeting_rules, created, active_from, active_to FROM campaigns WHERE channel_id = $1").await?;
+
+    let row = client.query(&statement, &[&campaign.channel.id()]).await?;
```
elpiel: you also have `query_one` which will return an error if the query returns 0 or >1 results.

sentry/src/routes/campaign.rs
```diff
@@ -20,25 +38,110 @@ pub async fn create_campaign<A: Adapter>(
...
+    match insert_or_modify_campaign(&app.pool, &campaign, &app.redis).await {
+        Err(error) => {
+            // error!(&app.logger, "{}", &error; "module" => "create_channel");
+            return Err(ResponseError::Conflict("channel already exists".to_string()));
+        }
+        Ok(false) => Err(error_response),
+        _ => Ok(()),
+    }?;
 
     let create_response = SuccessResponse { success: true };
```
elpiel: Should return the Campaign. At this point we have created the campaign ID randomly and we need to return the whole Campaign to caller. This way he will know the campaign Id

sentry/src/db/campaign.rs
simzzz: In this case, I'm trying to get all campaigns for the same channel with one query.

sentry/src/routes/campaign.rs
simzzz: How can we obtain this from Postgres?

elpiel: I will double check, but probably we would need to use analytics, since the spendable & last approved state are based on the Channel and not the Campaign.

PS: We haven't changed Analytics according to AIP#61 yet.

sentry/src/db/campaign.rs
elpiel: can you use a plural word - `rows`.

sentry/src/db/campaign.rs
```diff
@@ -49,6 +45,51 @@ pub async fn fetch_campaign(pool: DbPool, campaign: &Campaign) -> Result<Campaig
...
+
+    let campaigns = row.into_iter().map(|c| Campaign::from(&c)).collect();
```
elpiel: Should be: `rows.iter().map(Campaign::from).collect()`

`into_iter` will create an owned iterator (i.e. `|row: Row|` and for `Campaign::from` you need a `row: &Row` so no need for `into_iter()`, `iter()` is enough.

Since we call `iter()`, the `map()` gets a `row: &Row` which removes the need to wrap the `Campaign::from` in a closure all together, i.e. no need for `map(|row| Campaign::from(row))`

sentry/src/db/campaign.rs
```diff
@@ -49,6 +45,51 @@ pub async fn fetch_campaign(pool: DbPool, campaign: &Campaign) -> Result<Campaig
...
+        .prepare("SELECT EXISTS(SELECT 1 FROM campaigns WHERE id = $1)")
+        .await?;
+
+    let row = client.execute(&statement, &[&campaign.id]).await?;
+
+    let exists = row == 1;
+    Ok(exists)
+}
+
+pub async fn update_campaign_in_db(pool: &DbPool, campaign: &Campaign) -> Result<bool, PoolError> {
```
elpiel: `in_db` is redundant

sentry/src/db/campaign.rs
```diff
@@ -49,6 +45,51 @@ pub async fn fetch_campaign(pool: DbPool, campaign: &Campaign) -> Result<Campaig
...
+        .await?;
+
+    let row = client
+        .execute(&statement, &[
+            &campaign.budget,
+            &campaign.validators,
+            &campaign.title,
+            &campaign.pricing_bounds,
+            &campaign.event_submission,
+            &campaign.ad_units,
```
elpiel: Please make sure to test this function since `ad_units` might need to be wrapped in a `Json(campaign.ad_units)`

sentry/src/routes/campaign.rs
```diff
@@ -1,12 +1,38 @@
...
+use crate::db::campaign::{update_campaign_in_db, insert_campaign, get_campaigns_for_channel};
+use std::str::FromStr;
 
 pub async fn create_campaign<A: Adapter>(
     req: Request<Body>,
     app: &Application<A>,
 ) -> Result<Response<Body>, ResponseError> {
+    use serde::Serialize;
+    #[derive(Serialize)]
+    struct CreateCampaignResponse<'a> {
```
elpiel: You can directly return the `Campaign`. Now it will return:
```json
{
"campaign": { ..... }
}
```

sentry/src/routes/campaign.rs
elpiel: Skip this. If there is no record in Redis then something is gone terribly wrong.

sentry/src/db/campaign.rs
```diff
@@ -49,6 +45,59 @@ pub async fn fetch_campaign(pool: DbPool, campaign: &Campaign) -> Result<Campaig
     Ok(Campaign::from(&row))
 }
 
+pub async fn get_campaigns_for_channel(
+    pool: &DbPool,
+    campaign: &Campaign,
```
elpiel: - [x] Rename to `get_campaign_by_channel`
- [x] pass `ChannelId` instead of the whole `Campaign`

sentry/src/db/campaign.rs
```diff
@@ -49,6 +45,59 @@ pub async fn fetch_campaign(pool: DbPool, campaign: &Campaign) -> Result<Campaig
...
+    let statement = client.prepare("SELECT id, channel, creator, budget, validators, title, pricing_bounds, event_submission, ad_units, targeting_rules, created, active_from, active_to FROM campaigns WHERE channel_id = $1").await?;
+
+    let rows = client.query(&statement, &[&campaign.channel.id()]).await?;
+
+    let campaigns = rows.iter().map(Campaign::from).collect();
+
+    Ok(campaigns)
+}
+
+pub async fn campaign_exists(pool: &DbPool, campaign: &Campaign) -> Result<bool, PoolError> {
```
elpiel: Do we still need this fn?

sentry/src/db/campaign.rs
```diff
@@ -49,6 +45,59 @@ pub async fn fetch_campaign(pool: DbPool, campaign: &Campaign) -> Result<Campaig
...
+        .prepare("SELECT EXISTS(SELECT 1 FROM campaigns WHERE id = $1)")
+        .await?;
+
+    let row = client.execute(&statement, &[&campaign.id]).await?;
+
+    let exists = row == 1;
+    Ok(exists)
+}
+
+// TODO: Test for campaign ad_units
```
elpiel: - [ ] Still needs to be tested

sentry/src/routes/campaign.rs
```diff
@@ -17,28 +38,320 @@ pub async fn create_campaign<A: Adapter>(
...
+    let remaining_for_mutated_campaign = mutated_campaign.budget.checked_sub(&spent_for_mutated_campaign).ok_or_else(|| {
+        ResponseError::Conflict("Error while calculating remaining for mutated campaign".to_string())
+    })?;
+    sum_of_campaigns_remaining.checked_add(&remaining_for_mutated_campaign).ok_or_else(|| {
+        ResponseError::Conflict("Error while calculating sum for all campaigns".to_string())
+    })?;
+    Ok(sum_of_campaigns_remaining)
+}
+
+pub async fn modify_campaign(pool: &DbPool, campaign: &Campaign, redis: &MultiplexedConnection) -> Result<bool, ResponseError> {
```
elpiel: - [x] Should return the new Campaign
- [x] Return an error, not `ResponseError` (or create one if needed)
- [x] The function should receive the current Campaign & the modified fields (as not all fields are modifiable)
  We required this information, since we should calculate the difference between the NewBudget & the OldBudget and decide how the remaining should be updated in Redis (`INCRBY` or `DECRBY`)

sentry/src/routes/campaign.rs
```diff
@@ -17,28 +38,320 @@ pub async fn create_campaign<A: Adapter>(
...
+            _ => Ok(UnifiedNum::from_u64(0))
+        };
+    remaining
+}
+
+// tested
+async fn update_remaining_for_campaign(redis: &MultiplexedConnection, id: CampaignId, amount: UnifiedNum) -> Result<bool, ResponseError> {
+    // update a key in Redis for the remaining spendable amount
+    let key = format!("remaining:{}", id);
+    redis::cmd("INCRBY")
```
elpiel: This will only work if the Budget is increasing!
You must `INCRBY` or `DICRBY` accordingly to the difference.

Another way you can achieve this is to use `INCRBY -amount` but again - you must have this information available in this function, only `amount` is not enough.

sentry/src/routes/campaign.rs
```diff
@@ -17,28 +38,320 @@ pub async fn create_campaign<A: Adapter>(
...
+            },
+            _ => Ok(UnifiedNum::from_u64(0))
+        };
+    remaining
+}
+
+// tested
+async fn update_remaining_for_campaign(redis: &MultiplexedConnection, id: CampaignId, amount: UnifiedNum) -> Result<bool, ResponseError> {
+    // update a key in Redis for the remaining spendable amount
+    let key = format!("remaining:{}", id);
```
elpiel: - [x] add more context to the key - `campaignRemaining:{id}`
- [x] Extract the key `campaignRemaining` to a static, e.g. `static CAMPAIGN_REMAINING_KEY: &str = "campaignRemaining"`

sentry/src/routes/campaign.rs
```diff
@@ -17,28 +38,320 @@ pub async fn create_campaign<A: Adapter>(
...
+            return Err(ResponseError::Conflict("campaign already exists".to_string()));
+        }
+        Ok(false) => Err(error_response),
+        _ => Ok(()),
+    }?;
 
     Ok(success_response(serde_json::to_string(&campaign)?))
 }
+
+pub async fn update_campaign<A: Adapter>(
```
elpiel: This function can go to it's own module along with all the functions required for this route.
- [x] mod update_campaign { fn handle_route() }
- [x] move all functions & tests of the helper functions inside the module

sentry/src/routes/campaign.rs
```diff
@@ -17,28 +38,320 @@ pub async fn create_campaign<A: Adapter>(
...
+
+async fn get_campaigns_remaining_sum(redis: &MultiplexedConnection, campaigns: &[Campaign], mutated_campaign: &Campaign) -> Result<UnifiedNum, ResponseError> {
+    let other_campaigns_remaining = get_remaining_for_multiple_campaigns(&redis, &campaigns, mutated_campaign.id).await?;
+    let sum_of_campaigns_remaining = other_campaigns_remaining
+        .into_iter()
+        .try_fold(UnifiedNum::from_u64(0), |sum, val| sum.checked_add(&val).ok_or(ResponseError::Conflict("Couldn't sum remaining for campaigns".to_string())))?;
+    // Necessary to do it explicitly for current campaign as its budget is not yet updated in DB
+    let spent_for_mutated_campaign = get_spent_for_campaign(&redis, mutated_campaign.id).await?;
+    let remaining_for_mutated_campaign = mutated_campaign.budget.checked_sub(&spent_for_mutated_campaign).ok_or_else(|| {
+        ResponseError::Conflict("Error while calculating remaining for mutated campaign".to_string())
```
elpiel: The use of Conflict is only in places where we are trying to insert e.g. the same CampaignId into the database, i.e. the key/record already exists and there is a conflict in those primary keys in the DB.

sentry/src/routes/campaign.rs
```diff
@@ -17,28 +38,320 @@ pub async fn create_campaign<A: Adapter>(
...
+    })?;
+    Ok(total_remaining)
+}
+
+async fn get_remaining_for_multiple_campaigns(redis: &MultiplexedConnection, campaigns: &[Campaign], mutated_campaign_id: CampaignId) -> Result<Vec<UnifiedNum>, ResponseError> {
+    let other_campaigns_remaining = campaigns
+        .into_iter()
+        .filter(|c| c.id != mutated_campaign_id)
+        .map(|c| async move {
+            let spent = get_spent_for_campaign(&redis, c.id).await?;
```
elpiel: We should use the Redis `campaignRemaining` keys here.
I found that Redis has a special function for this `MGET` where you can define all the keys you want to fetch and it will return an Array with null-able (`nil`) values for those keys that were not found.

https://redis.io/commands/mget

sentry/src/routes/campaign.rs
```diff
@@ -17,28 +38,320 @@ pub async fn create_campaign<A: Adapter>(
...
+    Ok(sum_of_campaigns_remaining)
+}
+
+pub async fn modify_campaign(pool: &DbPool, campaign: &Campaign, redis: &MultiplexedConnection) -> Result<bool, ResponseError> {
+    let campaign_spent = get_spent_for_campaign(&redis, campaign.id).await?;
+    let accounting_spent = get_accounting_spent(pool.clone(), &campaign.creator, &campaign.channel.id()).await?;
+
+    let latest_spendable = fetch_spendable(pool.clone(), &campaign.creator, &campaign.channel.id()).await?;
+    // Check if we have reached the budget
+    if campaign_spent >= campaign.budget {
```
elpiel: If I understand correctly, since this is the NewBudget of the campaign - the check makes sense, however, the error message should be changed to reflect this.

sentry/src/routes/campaign.rs
```diff
@@ -1,7 +1,28 @@
...
+    adapter::Adapter,
+    sentry::{
+        campaign_create::CreateCampaign,
+    },
+    spender::Spendable,
+    Campaign, CampaignId, UnifiedNum, BigNum
+};
+use redis::aio::MultiplexedConnection;
+use slog::error;
+use crate::db::campaign::{update_campaign as update_campaign_db, insert_campaign, get_campaigns_for_channel};
```
elpiel: can you please leave the original name and instead rename the `fn update_campaign` to `mod modify_campaign { fn handle_route }`

primitives/src/sentry.rs
```diff
@@ -379,6 +381,18 @@ pub mod campaign_create {
             }
         }
     }
+
+    // All editable fields stored in one place, used for checking when a budget is changed
+    // #[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
+    // pub struct CampaignState {
```
elpiel: - [x] Use this struct for the modify Campaign route.
- [x] rename to `struct ModifyCampaign`, this way it will reflect more closely the use of the struct.

All fields of ModifyCampaign should be an `Option`

sentry/src/db/event_aggregate.rs
```diff
@@ -58,6 +59,24 @@ pub async fn latest_new_state(
...
+pub async fn latest_new_state_v5(
+    pool: &DbPool,
+    channel: &ChannelV5,
+    state_root: &str,
+) -> Result<Option<MessageResponse<NewState>>, PoolError> {
+    let client = pool.get().await?;
+
+    let select = client.prepare("SELECT \"from\", msg, received FROM validator_messages WHERE channel_id = $1 AND \"from\" = $2 AND msg ->> 'type' = 'NewState' AND msg->> 'stateRoot' = $3 ORDER BY received DESC LIMIT 1").await?;
+    let rows = client
+        .query(&select, &[&channel.id(), &channel.leader, &state_root])
```
elpiel: Older code is outdated this is why it doesn't use `query_opt` but newer can use it.

- [x] Use `query_opt()` instead

sentry/src/lib.rs
```diff
@@ -58,6 +60,8 @@ lazy_static! {
     static ref ADVERTISER_ANALYTICS_BY_CHANNEL_ID: Regex = Regex::new(r"^/analytics/for-advertiser/0x([a-zA-Z0-9]{64})/?$").expect("The regex should be valid");
     static ref PUBLISHER_ANALYTICS_BY_CHANNEL_ID: Regex = Regex::new(r"^/analytics/for-publisher/0x([a-zA-Z0-9]{64})/?$").expect("The regex should be valid");
     static ref CREATE_EVENTS_BY_CHANNEL_ID: Regex = Regex::new(r"^/channel/0x([a-zA-Z0-9]{64})/events/?$").expect("The regex should be valid");
+    static ref CAMPAIGN_UPDATE_BY_ID: Regex =
+        Regex::new(r"^/channel/0x([a-zA-Z0-9]{64})/?$").expect("The regex should be valid");
```
elpiel: it says `/channel`

- [x] fix typo `/channel` to `/campaign`
- [x] Use CampaignId regex: `0x([a-zA-Z0-9]{32})`

sentry/src/db/campaign.rs
```diff
@@ -72,7 +121,12 @@ mod test {
 
         assert!(is_inserted);
```
elpiel: Can you please add test against already existing Campaign? same as `is_inserted` but should error & matching the correct error: `PoolError::Backend(error) if error.code() = SqlState::UNIQUE_VIOLATION`

- [x] Add test for `Unique Violation` error

sentry/src/routes/campaign.rs
```diff
@@ -17,28 +38,320 @@ pub async fn create_campaign<A: Adapter>(
...
     Ok(success_response(serde_json::to_string(&campaign)?))
 }
+
+pub async fn update_campaign<A: Adapter>(
+    req: Request<Body>,
+    app: &Application<A>,
+) -> Result<Response<Body>, ResponseError> {
+    let body = hyper::body::to_bytes(req.into_body()).await?;
+
+    let campaign = serde_json::from_slice::<Campaign>(&body)
```
elpiel: - [x] use `struct ModifyCampaign`
- [x] get `Campaign` from `req.extensions()`
    use `.expect()` as we must have a `Campaign` in `extension()` or something terribly wrong has happened that can break the program.
   We insert this Campaign in `extension()` in the `fn campaigns_router`

sentry/src/routes/campaign.rs
```diff
@@ -17,28 +38,320 @@ pub async fn create_campaign<A: Adapter>(
...
-    //     }
-    //     Ok(false) => Err(error_response),
-    //     _ => Ok(()),
-    // }?;
-
-    let create_response = SuccessResponse { success: true };
+    match insert_campaign(&app.pool, &campaign).await {
+        Err(error) => {
+            error!(&app.logger, "{}", &error; "module" => "create_campaign");
+            return Err(ResponseError::Conflict("campaign already exists".to_string()));
```
elpiel: Only a single `error.code()` means that this campaign already exist:

`PoolError::Backend(error) if error.code() = SqlState::UNIQUE_VIOLATION`

The rest of the errors should be handled with `BadRequest` & the `error.to_string()`

- [x] check against UniqueViolation and return `Conflict` response
- [x] `Ok(false)` - stays the same, just use `=> Err(ResponseError::BadRequest("Encountered error while creating Campaign; please try again".to_string()))`

sentry/src/lib.rs
```diff
@@ -308,6 +324,28 @@ async fn channels_router<A: Adapter + 'static>(
...
+    mut req: Request<Body>,
+    app: &Application<A>,
+) -> Result<Response<Body>, ResponseError> {
+    req = AuthRequired.call(req, app).await?;
+
+    let (path, method) = (req.uri().path().to_owned(), req.method());
+
+    // regex matching for routes with params
+    if let (Some(caps), &Method::POST) = (CAMPAIGN_UPDATE_BY_ID.captures(&path), method) {
+        let param = RouteParams(vec![caps
```
elpiel: Here there are a few things we can make:
- [x] Use the parameter and parse the `param.get(0).ok_or_else(BadRequest("No CampaignId")`
- [x] Fetch `Campaign` using the parsed `CampaignId` ( None should be handled as ResponseError::NotFound)
- [x] Add `Campaign` to `req.extensions_mut()` (see a few lines bellow)

sentry/src/routes/campaign.rs
```diff
@@ -17,28 +38,320 @@ pub async fn create_campaign<A: Adapter>(
...
+        }
+        Ok(false) => Err(error_response),
+        _ => Ok(()),
+    }?;
+
+    Ok(success_response(serde_json::to_string(&campaign)?))
+}
+
+// TODO: Double check redis calls
+async fn get_spent_for_campaign(redis: &MultiplexedConnection, id: CampaignId) -> Result<UnifiedNum, ResponseError> {
```
elpiel: We should not use a `spent` key. The Spent can be calculated by `campaign.budget - campaignRemaining` if we need it.

- [x] Remove function

sentry/src/routes/campaign.rs
```diff
@@ -17,28 +38,320 @@ pub async fn create_campaign<A: Adapter>(
...
+                Ok(UnifiedNum::from_u64(res))
+            },
+            _ => Ok(UnifiedNum::from_u64(0))
+        };
+
+    campaign_spent
+}
+
+async fn get_remaining_for_campaign_from_redis(redis: &MultiplexedConnection, id: CampaignId) -> Result<UnifiedNum, ResponseError> {
+    let key = format!("remaining:{}", id);
```
elpiel: The Key `campaignRemaining` instead of `remaining` only goes to a static variable in the new `mod modify_campaign`

sentry/src/routes/campaign.rs
```diff
@@ -17,28 +38,320 @@ pub async fn create_campaign<A: Adapter>(
...
+    let remaining = match redis::cmd("GET")
+        .arg(&key)
+        .query_async::<_, Option<String>>(&mut redis.clone())
+        .await
+         {
+            Ok(Some(remaining)) => {
```
elpiel: Leave a todo for this as we are waiting for #412 to get merged.
- [x] Leave a todo to use `FromStr` (i.e. `.parse()`) to properly parse the `UnifiedNum` from a string redis value. Mention PR #412 & issue #382 and leave an `to_u64().unwrap()`
- [x] Add to tasks in issue https://github.com/AdExNetwork/adex-validator-stack-rust/issues/382

sentry/src/routes/campaign.rs
```diff
@@ -17,28 +38,320 @@ pub async fn create_campaign<A: Adapter>(
...
+                })?;
+                Ok(UnifiedNum::from_u64(res))
+            },
+            _ => Ok(UnifiedNum::from_u64(0))
+        };
+
+    campaign_spent
+}
+
+async fn get_remaining_for_campaign_from_redis(redis: &MultiplexedConnection, id: CampaignId) -> Result<UnifiedNum, ResponseError> {
```
elpiel: - [x] Do not return a `ResponseError` here.

sentry/src/routes/campaign.rs
```diff
@@ -17,28 +38,320 @@ pub async fn create_campaign<A: Adapter>(
...
+        .await
+         {
+            Ok(Some(remaining)) => {
+                let res = BigNum::from_str(&remaining)?;
+                let res = res.to_u64().ok_or_else(|| {
+                    ResponseError::Conflict("Error while calculating the total remaining amount".to_string())
+                })?;
+                Ok(UnifiedNum::from_u64(res))
+            },
+            _ => Ok(UnifiedNum::from_u64(0))
```
elpiel: - [ ] Return proper error
- [ ] Use Option and do return `None` and not `0` as default

sentry/src/routes/campaign.rs
```diff
@@ -17,28 +38,320 @@ pub async fn create_campaign<A: Adapter>(
...
+        };
+    remaining
+}
+
+// tested
+async fn update_remaining_for_campaign(redis: &MultiplexedConnection, id: CampaignId, amount: UnifiedNum) -> Result<bool, ResponseError> {
```
elpiel: - [x] Use proper error instead of `ResponseError` here

elpiel (changes requested):

sentry/src/db/campaign.rs
```diff
@@ -50,11 +48,56 @@ pub async fn fetch_campaign(
...
+}
+
+// TODO: Test for campaign ad_units
+pub async fn update_campaign(pool: &DbPool, campaign: &Campaign) -> Result<bool, PoolError> {
+    let client = pool.get().await?;
+    let statement = client
+        .prepare("UPDATE campaigns SET budget = $1, validators = $2, title = $3, pricing_bounds = $4, event_submission = $5, ad_units = $6, targeting_rules = $7 WHERE id = $8")
+        .await?;
+
+    let row = client
```
elpiel: execute returns number of rows, not a `Row` object, maybe a better fit naming could be used here.

sentry/src/db/campaign.rs
```diff
@@ -80,6 +123,17 @@ mod test {
...
+        let is_duplicate_inserted = insert_campaign(&database.pool, &campaign_for_testing).await;
+
+        assert!(is_duplicate_inserted.is_err());
+        let insertion_error = is_duplicate_inserted.err().expect("should get error");
+        match insertion_error {
+            PoolError::Backend(error) if error.code() == Some(&SqlState::UNIQUE_VIOLATION) => {
+                assert!(true);
+            }
+            _ => assert!(false),
+        }
```
elpiel: Instead of expecting and checking `is_err()`, just use `matches!()` to match against the `PoolError` & `error.code()`
https://doc.rust-lang.org/std/macro.matches.html#examples

sentry/src/db/campaign.rs
elpiel: I don't see any tests for this function, could you please add them?

sentry/src/lib.rs
```diff
@@ -180,7 +195,27 @@ async fn campaigns_router<A: Adapter + 'static>(
     let (path, method) = (req.uri().path(), req.method());
 
     // create events
```
elpiel: this comment is not suitable anymore.

sentry/src/lib.rs
```diff
@@ -61,6 +63,8 @@ lazy_static! {
     static ref ADVERTISER_ANALYTICS_BY_CHANNEL_ID: Regex = Regex::new(r"^/analytics/for-advertiser/0x([a-zA-Z0-9]{64})/?$").expect("The regex should be valid");
     static ref PUBLISHER_ANALYTICS_BY_CHANNEL_ID: Regex = Regex::new(r"^/analytics/for-publisher/0x([a-zA-Z0-9]{64})/?$").expect("The regex should be valid");
     static ref CREATE_EVENTS_BY_CHANNEL_ID: Regex = Regex::new(r"^/channel/0x([a-zA-Z0-9]{64})/events/?$").expect("The regex should be valid");
+    static ref CAMPAIGN_UPDATE_BY_ID: Regex =
```
elpiel: could you please move this Regex to use the way bellow (without a macro) with Lazy.

sentry/src/lib.rs
```diff
@@ -180,7 +195,27 @@ async fn campaigns_router<A: Adapter + 'static>(
...
+            .ok_or_else(|| ResponseError::BadRequest("No CampaignId".to_string()))?;
+        let campaign_id = CampaignId::from_str(&campaign_id)
+            .map_err(|_| ResponseError::BadRequest("Bad CampaignId".to_string()))?;
+
+        let campaign = fetch_campaign(app.pool.clone(), &campaign_id)
+            .await
+            .map_err(|_| ResponseError::NotFound)?;
+
+        req.extensions_mut().insert(campaign);
+        handle_route(req, app).await
```
elpiel: Can you add the module name in front instead of importing the function directly? This way it will be more clear and a generic name as `handle_route` can be given context (and be used in other modules for example) as well.

sentry/src/routes/campaign.rs
```diff
@@ -25,32 +68,197 @@ pub async fn create_campaign<A: Adapter>(
 
     // TODO AIP#61: Validate Campaign
 
-    let error_response = ResponseError::BadRequest("err occurred; please try again later".into());
+    let error_response = ResponseError::BadRequest("err occurred; please try again later".to_string());
 
-    // insert Campaign
+    let accounting_spent = get_accounting_spent(app.pool.clone(), &campaign.creator, &campaign.channel.id()).await?;
```
elpiel: At this point, the `AuthRequired` middleware has verified that the Auth token is valid and allowed Address in the config (FYI: the type should be altered to Address not ValidatorId). We still need, however, to validate that the Sender of the request is the `campaign.creator`, as they must be the same.

- [x] Validate that the Request sender (from Session) is the `campaign.creator` when handling campaign creation.

sentry/src/routes/campaign.rs
```diff
@@ -25,32 +68,197 @@ pub async fn create_campaign<A: Adapter>(
 
     // TODO AIP#61: Validate Campaign
```
elpiel: Please create a Campaign from the CreateCampaign (it will only generate a random ID) and validate the resulting Campaign.

- [x] validate the `Campaign` by using `campaign_validator::Validator` trait

sentry/src/routes/campaign.rs
```diff
@@ -25,32 +68,197 @@ pub async fn create_campaign<A: Adapter>(
...
-    //     Ok(false) => Err(error_response),
-    //     _ => Ok(()),
-    // }?;
-
-    let create_response = SuccessResponse { success: true };
+    if campaign.budget > remaining_for_channel {
+        return Err(ResponseError::BadRequest("Not Enough budget for campaign".to_string()));
+    }
+
+    // If the channel is being created, the amount spent is 0, therefore remaining = budget
```
elpiel: - [x] should say `Campaign`

sentry/src/routes/campaign.rs
```diff
@@ -25,32 +68,197 @@ pub async fn create_campaign<A: Adapter>(
...
+
+    let total_remaining = total_deposited.checked_sub(&accounting_spent);
+    total_remaining
+}
+
+pub mod update_campaign {
+    use super::*;
+    use lazy_static::lazy_static;
+
+    lazy_static!{
```
elpiel: Since you don't need allocation of memory (i.e. you don't use a Vec, HashMap, etc.) you can remove the `lazy_static`. Since we have `&'static str` the compiler can directly embed it into the binary of the program.

Same goes for `const fn` or stucts that don't require some sort of allocation of memory. Having `Vec::new()` (empty vec) is allowed as it doesn't allocate new memory but `vec.push()` is not.

https://doc.rust-lang.org/std/keyword.static.html

sentry/src/routes/campaign.rs
```diff
@@ -25,32 +68,197 @@ pub async fn create_campaign<A: Adapter>(
...
+
+        Ok(success_response(serde_json::to_string(&campaign)?))
+    }
+
+    pub async fn modify_campaign(pool: &DbPool, campaign: &Campaign, modified_campaign: &ModifyCampaign, redis: &MultiplexedConnection) -> Result<Campaign, CampaignError> {
+        // *NOTE*: When updating campaigns make sure sum(campaigns.map(getRemaining)) <= totalDepoisted - totalspent
+        // !WARNING!: totalSpent != sum(campaign.map(c => c.spending)) therefore we must always calculate remaining funds based on total_deposit - lastApprovedNewState.spenders[user]
+        // *NOTE*: To close a campaign set campaignBudget to campaignSpent so that spendable == 0
+
+        let new_budget = modified_campaign.budget.ok_or(CampaignError::FailedUpdate("Couldn't get new budget".to_string()))?;
```
elpiel: It's not mandatory to update the budget when you send a Modify Campaign request. This should **not** fail and instead handle the new Budget if it's `Some`
- [x] Handle budget only if it's passed through the request

sentry/src/routes/campaign.rs
```diff
@@ -25,32 +68,197 @@ pub async fn create_campaign<A: Adapter>(
...
+        }
+        Ok(false) => Err(ResponseError::BadRequest("Encountered error while creating Campaign; please try again".to_string())),
+        _ => Ok(()),
+    }?;
 
     Ok(success_response(serde_json::to_string(&campaign)?))
 }
 
+
+// tested
```
elpiel: not entirely sure if we need a whole function for this.
It only performs single line `latest_spendable.deposit.total.checked_sub(accounting_spent)`

sentry/src/routes/campaign.rs
```diff
@@ -25,32 +68,197 @@ pub async fn create_campaign<A: Adapter>(
...
+
+pub mod update_campaign {
+    use super::*;
+    use lazy_static::lazy_static;
+
+    lazy_static!{
+        pub static ref CAMPAIGN_REMAINING_KEY: &'static str = "campaignRemaining";
+    }
+
+    pub async fn increase_remaining_for_campaign(redis: &MultiplexedConnection, id: CampaignId, amount: UnifiedNum) -> Result<bool, CampaignError> {
```
elpiel: I'm hesitant if we should use `INCRBY` for `Campaign` creation. It makes more sense to fail if the campaign Id is already in Redis.

We can use the `SET` with the option `NX -- Only set the key if it does not already exist` and return an error if there is value already for this `CampaignId`.

https://redis.io/commands/set

How does it sound to you @simzzz ?

sentry/src/routes/campaign.rs
```diff
@@ -25,32 +68,197 @@ pub async fn create_campaign<A: Adapter>(
...
+    }
+
+    pub async fn decrease_remaining_for_campaign(redis: &MultiplexedConnection, id: CampaignId, amount: UnifiedNum) -> Result<bool, CampaignError> {
+        let key = format!("{}:{}", *CAMPAIGN_REMAINING_KEY, id);
+        redis::cmd("DECRBY")
+            .arg(&key)
+            .arg(amount.to_u64())
+            .query_async(&mut redis.clone())
+            .await?;
+        Ok(true)
```
elpiel: The redis command will return the new value. We must check if this value is still positive when modifying the Campaign Budget!

- [x] Return the value from the redis commmand
- [x] Write a validation that it is still `> 0` when modifying campaign

elpiel (changes requested):

sentry/src/routes/campaign.rs
```diff
@@ -25,32 +68,197 @@ pub async fn create_campaign<A: Adapter>(
...
+        let latest_spendable = fetch_spendable(pool.clone(), &campaign.creator, &campaign.channel.id()).await?;
+
+        let old_remaining = get_remaining_for_campaign_from_redis(&redis, campaign.id).await.ok_or(CampaignError::FailedUpdate("Couldn't get remaining for campaign".to_string()))?;
+
+        let campaign_spent = new_budget.checked_sub(&old_remaining).ok_or(CampaignError::CalculationError)?;
+        if campaign_spent >= new_budget {
+            return Err(CampaignError::BudgetExceeded);
+        }
+
+        let old_remaining = UnifiedNum::from_u64(max(0, old_remaining.to_u64()));
```
elpiel: You can use `max()` with any type that impl `Ord`, `UnifiedNum` does, so no need to  do conversation between `u64` & `UnifiedNum`, just use `UnifiedNum`.

sentry/src/routes/campaign.rs
```diff
@@ -25,32 +68,197 @@ pub async fn create_campaign<A: Adapter>(
...
+        // *NOTE*: To close a campaign set campaignBudget to campaignSpent so that spendable == 0
+
+        let new_budget = modified_campaign.budget.ok_or(CampaignError::FailedUpdate("Couldn't get new budget".to_string()))?;
+        let accounting_spent = get_accounting_spent(pool.clone(), &campaign.creator, &campaign.channel.id()).await?;
+
+        let latest_spendable = fetch_spendable(pool.clone(), &campaign.creator, &campaign.channel.id()).await?;
+
+        let old_remaining = get_remaining_for_campaign_from_redis(&redis, campaign.id).await.ok_or(CampaignError::FailedUpdate("Couldn't get remaining for campaign".to_string()))?;
+
+        let campaign_spent = new_budget.checked_sub(&old_remaining).ok_or(CampaignError::CalculationError)?;
```
elpiel: We need to calculate the Spent amount based on the Old budget, before modifying the campaign with new budget!

`campaign_spent = old_budget - remaining`.
Return error if None is returned as you do here
- [x] Calculation should be based on the Old Budget

sentry/src/routes/campaign.rs
```diff
@@ -25,32 +68,197 @@ pub async fn create_campaign<A: Adapter>(
...
+        if campaign_spent >= new_budget {
+            return Err(CampaignError::BudgetExceeded);
```
elpiel: Error should be more clear here. Spent amount is greater than the new budget of campaign.
e.g. `CampaignError::NewBudget("New budget should be greater than the spent amount")`

sentry/src/routes/campaign.rs
```diff
@@ -1,16 +1,59 @@
...
+    str::FromStr,
+    collections::HashMap,
+};
+use deadpool_postgres::PoolError;
+use tokio_postgres::error::SqlState;
+
+use chrono::Utc;
+
+#[derive(Debug, PartialEq, Eq)]
+pub enum CampaignError {
```
elpiel: `Error` should be enough for now and there is no need to suffix the `Calculation` error with `Error`.

- [x] rename CampaignError struct to `Error`
- [x] remove suffix of `CalculationError`
- [x] `derive(Error)` (thiserrror::Error)
- [x] Add Redis & Postgres cases (e.g. `sentry::db::RedisError` & `sentry::db::PoolError`)
- [x] Add `#[from]`(from `thiserror` crate) for Postgres & Redis errors instead of `impl From<...> for Error`

sentry/src/routes/campaign.rs
elpiel: Please put this logic in the `campaign_validator::Validator` and pass a `for_creator: Address` which will come from the request `Session`.

sentry/src/routes/campaign.rs
```diff
@@ -25,32 +68,197 @@ pub async fn create_campaign<A: Adapter>(
...
+            return Err(CampaignError::BudgetExceeded);
+        }
+
+        update_campaign(&pool, &campaign).await?;
+
+        Ok(campaign.clone())
+    }
+
+    // TODO: #382 Remove after #412 is merged
+    fn get_unified_num_from_string(value: &str) -> Option<UnifiedNum> {
```
elpiel: You can use the `FromStr` impl of `UnifiedNum` i.e. `.parse()`

- [x] Remove this function and use `.parse()` instead.

sentry/src/routes/campaign.rs
```diff
@@ -25,32 +68,197 @@ pub async fn create_campaign<A: Adapter>(
...
+            _ => None,
+        };
+        let value_as_unified = match value_as_u64 {
+            Some(num) => Some(UnifiedNum::from_u64(num)),
+            _ => None
+        };
+        value_as_unified
+    }
+
+    pub async fn get_remaining_for_campaign_from_redis(redis: &MultiplexedConnection, id: CampaignId) -> Option<UnifiedNum> {
```
elpiel: - [x] remove `_from_redis`
- [x] Should return RedisError and the `.parse()` error (you can use the currently named `CampaignError` or `Error` when it gets renamed)

I'm also not 100% sure that we can use `GET` only, but will check the logic later.

sentry/src/routes/campaign.rs
```diff
@@ -25,32 +68,197 @@ pub async fn create_campaign<A: Adapter>(
...
+        let key = format!("{}:{}", *CAMPAIGN_REMAINING_KEY, id);
+        let remaining = match redis::cmd("GET")
+            .arg(&key)
+            .query_async::<_, Option<String>>(&mut redis.clone())
+            .await
+            {
+                Ok(Some(remaining)) => {
```
elpiel: There is a chance for this value to be **Negative** which the UnifiedNum does **not** support, return `0` (`UnifiedNum::zero()`) instead.

sentry/src/routes/campaign.rs
```diff
@@ -25,32 +68,197 @@ pub async fn create_campaign<A: Adapter>(
...
+                    // TODO: #382 Just parse from string once #412 is merged
+                    get_unified_num_from_string(&remaining)
+                },
+                _ => None
+            };
+        remaining
+    }
+
+    async fn get_remaining_for_multiple_campaigns(redis: &MultiplexedConnection, campaigns: &[Campaign]) -> Result<Vec<UnifiedNum>, CampaignError> {
+        let keys: Vec<String> = campaigns.into_iter().map(|c| format!("{}:{}", *CAMPAIGN_REMAINING_KEY, c.id)).collect();
```
elpiel: `iter()` should be enough

sentry/src/routes/campaign.rs
```diff
@@ -25,32 +68,197 @@ pub async fn create_campaign<A: Adapter>(
...
+        let remainings = redis::cmd("MGET")
+            .arg(keys)
+            .query_async::<_, Vec<Option<String>>>(&mut redis.clone())
+            .await?;
+
+        let remainings = remainings
```
elpiel: We should double check how to handle the `None` cases.
These `flat_map().map().flatten()` are quite cumbersome and should be removed. Use `0` (`UnifiedNum::zero()`) when None is returned, i.e. when the campaign remaining is not found in Redis.

sentry/src/routes/campaign.rs
```diff
@@ -25,32 +68,197 @@ pub async fn create_campaign<A: Adapter>(
...
+            .collect();
+
+        Ok(remainings)
+    }
+
+    pub async fn get_campaigns_remaining_sum(redis: &MultiplexedConnection, campaigns: &[Campaign], mutated_campaign: CampaignId, new_budget: &UnifiedNum) -> Result<UnifiedNum, CampaignError> {
+        let other_campaigns_remaining = get_remaining_for_multiple_campaigns(&redis, &campaigns).await?;
+        let sum_of_campaigns_remaining = other_campaigns_remaining
+            .into_iter()
+            .try_fold(UnifiedNum::from_u64(0), |sum, val| sum.checked_add(&val).ok_or(CampaignError::CalculationError))?;
```
elpiel: For an Iterator of `UnifiedNum`s you can use `sum()` with return type `Option<UnifiedNum>`. There is no need for `try_fold` and so on.

None is returned when overflow happens. I think I've mentioned this before. Check the `UnifiedNum` implementations.

sentry/src/routes/campaign.rs
```diff
@@ -25,32 +68,197 @@ pub async fn create_campaign<A: Adapter>(
...
+
+        // Necessary to do it explicitly for current campaign as its budget is not yet updated in DB
+        let old_remaining_for_mutated_campaign = get_remaining_for_campaign_from_redis(&redis, mutated_campaign).await.ok_or(CampaignError::CalculationError)?;
+        let spent_for_mutated_campaign = new_budget.checked_sub(&old_remaining_for_mutated_campaign).ok_or(CampaignError::CalculationError)?;
+        let new_remaining_for_mutated_campaign = new_budget.checked_sub(&spent_for_mutated_campaign).ok_or(CampaignError::CalculationError)?;
+        sum_of_campaigns_remaining.checked_add(&new_remaining_for_mutated_campaign).ok_or(CampaignError::CalculationError)?;
+        Ok(sum_of_campaigns_remaining)
```
elpiel: This whole part looks quite smelly. The original solution:

> key = remaining:{campaignId}
> oldRemaining = redis.get(key)
// This part will be handled by the check if the value returned from Redis is negative to use `UnifiedNum::zero()`
> remaining = max(0, oldRemaining)

You must use `newBudget.cmp(oldBudget)` and depending on https://doc.rust-lang.org/std/cmp/enum.Ordering.html

You should use:

- `Ordering::Greater`

> newRemaining = oldRemaining + (newBudget - oldBudget)
> redis.decrby(key, newRemaining - oldRemaining)
e.g. 100 + (500 - 400), newRemaining = 200

(200 - 100) = `redis incrby 100`

- `Ordering::Less`

> newRemaining = oldRemaining + (oldBudget - newBudget)
> redis.incrby(key, newRemaining - oldRemaining)

e.g. 100 - (400 - 350), newRemaining = 50

(100 - 50) = `redis decrby 50`

sentry/src/routes/campaign.rs
```diff
@@ -123,3 +331,130 @@ async fn process_events<A: Adapter + 'static>(
...
+    use super::*;
+
+    async fn get_redis() -> Object<Manager> {
+        let connection = TESTS_POOL.get().await.expect("Should return Object");
+        connection
+    }
+
+    fn get_campaign() -> Campaign {
+        DUMMY_CAMPAIGN.clone()
+    }
```
elpiel: single line functions could be remove

sentry/src/routes/campaign.rs
```diff
@@ -123,3 +331,130 @@ async fn process_events<A: Adapter + 'static>(
...
+
+    fn get_dummy_spendable(spender: Address) -> Spendable {
```
elpiel: since this is used for the campaign, it's better to pass the Channel or the Campaign itself.
If we use a different campaign in the test with different Channel, this function might use a different channel (because of the `DUMMY_CAMPAIGN.channel.clone(),`) and can cause tests to fail.

sentry/src/routes/campaign.rs
simzzz: Yes, that would be a better solution.

sentry/src/routes/campaign.rs
```diff
@@ -123,3 +331,130 @@ async fn process_events<A: Adapter + 'static>(
...
+
+        let remaining = redis::cmd("GET")
+            .arg(&key)
+            .query_async::<_, Option<String>>(&mut redis.connection)
+            .await
+            .expect("should get remaining");
+        assert_eq!(remaining.is_some(), true);
+
+        let remaining = remaining.expect("should get remaining");
+        let remaining = UnifiedNum::from_u64(remaining.parse::<u64>().expect("should parse"));
```
elpiel: You can directly parse the string into `UnifiedNum`. Also you can simplify these lines and just have 1 `remaining` var.

sentry/src/routes/campaign.rs
```diff
@@ -123,3 +331,130 @@ async fn process_events<A: Adapter + 'static>(
...
+        let should_be_remaining = UnifiedNum::from_u64(500).checked_add(&campaign.budget).expect("should add");
```
elpiel: You can use `checked_add` but this is a test and we don't really need to handle overflow here.

sentry/src/routes/campaign.rs
```diff
@@ -123,3 +331,130 @@ async fn process_events<A: Adapter + 'static>(
...
+            .expect("should get remaining");
+        assert_eq!(remaining.is_some(), true);
+
+        let remaining = remaining.expect("should get remaining");
+        let remaining = UnifiedNum::from_u64(remaining.parse::<u64>().expect("should parse"));
+        assert_eq!(remaining, should_be_remaining);
+    }
+
+    #[tokio::test]
+    async fn update_remaining_before_it_is_set() {
```
elpiel: You are testing if Redis will return `None` on a key that was not set.
There is no actual test here, you can remove it.

- [x] remove test

sentry/src/db/campaign.rs
```diff
@@ -50,11 +48,90 @@ pub async fn fetch_campaign(
...
+    let statement = client.prepare("SELECT id, channel, creator, budget, validators, title, pricing_bounds, event_submission, ad_units, targeting_rules, created, active_from, active_to FROM campaigns WHERE channel_id = $1").await?;
+
+    let rows = client.query(&statement, &[&channel_id]).await?;
+
+    let campaigns = rows.iter().map(Campaign::from).collect();
+
+    Ok(campaigns)
+}
+
+pub async fn update_campaign(
```
elpiel: Instead of passing both, we can just pass the Campaign and instead have a method on `ModifyCampaign` to `apply()` the changes to a `Campaign`. This way the database call function will be strictly for updating the database and not handling any additional logic regarding modifying a campaign (and will be reusable in other scenarios)

sentry/src/db/campaign.rs
```diff
@@ -50,11 +48,90 @@ pub async fn fetch_campaign(
     Ok(row.as_ref().map(Campaign::from))
 }
 
+pub async fn get_campaigns_by_channel(
```
elpiel, line 62: We have to see how we will paginate the results. Postgres could block if there are many campaigns for a single channel.

We might need to use a `LIMIT` query multiple times to fetch all campaigns.
Please add a TODO and a Task in the issue #409

sentry/src/lib.rs
```diff
@@ -179,8 +195,27 @@ async fn campaigns_router<A: Adapter + 'static>(
...
+        // Should be safe to access indice here
+        let campaign_id = param
+            .get(0)
+            .ok_or_else(|| ResponseError::BadRequest("No CampaignId".to_string()))?;
+        let campaign_id = CampaignId::from_str(&campaign_id)
+            .map_err(|_| ResponseError::BadRequest("Bad CampaignId".to_string()))?;
+
+        let campaign = fetch_campaign(app.pool.clone(), &campaign_id)
+            .await
+            .map_err(|_| ResponseError::NotFound)?;
```
elpiel: You can use the `CampaignLoad` middleware

sentry/src/lib.rs
```diff
@@ -69,6 +71,9 @@ static INSERT_EVENTS_BY_CAMPAIGN_ID: Lazy<Regex> = Lazy::new(|| {
 static CLOSE_CAMPAIGN_BY_CAMPAIGN_ID: Lazy<Regex> = Lazy::new(|| {
     Regex::new(r"^/campaign/0x([a-zA-Z0-9]{32})/close/?$").expect("The regex should be valid")
 });
+static CAMPAIGN_UPDATE_BY_ID: Lazy<Regex> = Lazy::new(|| {
```
elpiel, line 74: - [ ] Please prefix the route with `/v5/`

sentry/src/lib.rs
```diff
@@ -69,6 +71,9 @@ static INSERT_EVENTS_BY_CAMPAIGN_ID: Lazy<Regex> = Lazy::new(|| {
 static CLOSE_CAMPAIGN_BY_CAMPAIGN_ID: Lazy<Regex> = Lazy::new(|| {
```
elpiel, line 71: - [ ] Please prefix the route with `/v5/`

sentry/src/lib.rs
```diff
@@ -158,6 +163,17 @@ impl<A: Adapter + 'static> Application<A> {
 
                 publisher_analytics(req, &self).await
             }
+            // For creating campaigns
+            ("/campaign", &Method::POST) => {
```
elpiel: - [ ] Please prefix the route with `/v5/`

sentry/src/routes/campaign.rs
```diff
@@ -1,56 +1,286 @@
...
+    Redis(#[from] RedisError),
+    #[error("DB Pool error: {0}")]
+    Pool(#[from] PoolError)
+}
 
 pub async fn create_campaign<A: Adapter>(
     req: Request<Body>,
     app: &Application<A>,
 ) -> Result<Response<Body>, ResponseError> {
+    let session = req
```
elpiel: Session & Auth are 2 separate things. Please make sure to
- [x] rename correctly variable to struct.

sentry/src/routes/campaign.rs
```diff
@@ -1,56 +1,286 @@
...
         .map_err(|e| ResponseError::FailedValidation(e.to_string()))?
         // create the actual `Campaign` with random `CampaignId`
         .into_campaign();
 
-    // TODO AIP#61: Validate Campaign
+    campaign.validate(&app.config, &app.adapter.whoami()).map_err(|_| ResponseError::FailedValidation("couldn't valdiate campaign".to_string()))?;
 
-    let error_response = ResponseError::BadRequest("err occurred; please try again later".into());
+    // TODO: Just use session.uid once it's address
+    if Address::from_bytes(session.uid.as_bytes()) != campaign.creator {
```
elpiel: - [x] `ValidatorId` has a method `to_address()`

sentry/src/routes/campaign.rs
```diff
@@ -1,56 +1,286 @@
...
-    //             }
-    //             _ => Err(error_response),
-    //         }
-    //     }
-    //     Ok(false) => Err(error_response),
-    //     _ => Ok(()),
-    // }?;
-
-    let create_response = SuccessResponse { success: true };
+    let remaining_for_channel = total_deposited.checked_sub(&accounting_spent).ok_or(ResponseError::FailedValidation("couldn't calculate remaining for channel".to_string()))?;
```
elpiel: `No more budget` would be more appropriate

sentry/src/routes/campaign.rs
```diff
@@ -1,56 +1,286 @@
...
+
+    if campaign.budget > remaining_for_channel {
+        return Err(ResponseError::BadRequest("Not Enough budget for campaign".to_string()));
```
elpiel: - [ ] `Not enough deposit left for the new campaign budget`

sentry/src/routes/campaign.rs
```diff
@@ -1,56 +1,286 @@
...
     Ok(success_response(serde_json::to_string(&campaign)?))
 }
 
+pub mod update_campaign {
+    use super::*;
+    use crate::fetch_campaign;
+
+    pub const CAMPAIGN_REMAINING_KEY: &'static str = "campaignRemaining";
+
+    pub async fn set_initial_remaining_for_campaign(redis: &MultiplexedConnection, id: CampaignId, amount: UnifiedNum) -> Result<bool, Error> {
```
elpiel: Return a value that would be used to check if the Campaign id already existed in Redis. This would be another check if the campaign already exist in combination with the Postgres one.

sentry/src/routes/campaign.rs
```diff
@@ -1,56 +1,286 @@
...
+        let key = format!("{}:{}", CAMPAIGN_REMAINING_KEY, id);
+        redis::cmd("SETNX")
+            .arg(&key)
+            .arg(amount.to_u64())
+            .query_async(&mut redis.clone())
```
elpiel: No need to clone everywhere. Use e.g. `redis: &mut MultiplexedConnection` instead

sentry/src/routes/campaign.rs
```diff
@@ -1,56 +1,286 @@
...
+            .await?;
+        Ok(true)
+    }
+
+    pub async fn decrease_remaining_for_campaign(redis: &MultiplexedConnection, id: CampaignId, amount: UnifiedNum) -> Option<UnifiedNum> {
+        let key = format!("{}:{}", CAMPAIGN_REMAINING_KEY, id);
+        let value = match redis::cmd("DECRBY")
+            .arg(&key)
+            .arg(amount.to_u64())
+            .query_async::<_, Option<i64>>(&mut redis.clone())
```
elpiel: same goes for this `clone()`

sentry/src/routes/campaign.rs
```diff
@@ -1,56 +1,286 @@
...
+            };
+        value
+    }
+
+    pub async fn handle_route<A: Adapter>(
+        req: Request<Body>,
+        app: &Application<A>,
+    ) -> Result<Response<Body>, ResponseError> {
+        let campaign = req.extensions().get::<Campaign>().expect("We must have a campaign in extensions");
+        let modified_campaign = ModifyCampaign::from_campaign(campaign.clone());
```
elpiel: - [x] `MofieidCampaign` should be created from the request body.
Later on we can `modify_campaign.apply(&mut campaign)` to apply the changes from the ModifiedCampaign => Campaign.

sentry/src/routes/campaign.rs
```diff
@@ -1,56 +1,286 @@
...
+
+        // Getting the campaign as it was before the update operation, should exist
```
elpiel: You already have this value, it's in the `req.extensions()`

sentry/src/routes/campaign.rs
```diff
@@ -1,56 +1,286 @@
...
+            let campaign_spent = campaign.budget.checked_sub(&old_remaining).ok_or(Error::Calculation)?;
+            if campaign_spent >= new_budget {
+                return Err(Error::NewBudget("New budget should be greater than the spent amount".to_string()));
+            }
+
+            // Separate variable for clarity
+            let old_budget = campaign.budget;
+
+            match new_budget.cmp(&old_budget) {
+                Ordering::Greater | Ordering::Equal => {
```
elpiel: if New & Old budgets are equal you can skip this step of calculating and calling redis.

sentry/src/routes/campaign.rs
```diff
@@ -1,56 +1,286 @@
...
+                    let amount_to_incr = new_remaining.checked_sub(&old_remaining).ok_or(Error::Calculation)?;
+                    increase_remaining_for_campaign(&redis, campaign.id, amount_to_incr).await?;
+                },
+                Ordering::Less => {
+                    let new_remaining = old_remaining.checked_add(&old_budget.checked_sub(&new_budget).ok_or(Error::Calculation)?).ok_or(Error::Calculation)?;
+                    let amount_to_decr = new_remaining.checked_sub(&old_remaining).ok_or(Error::Calculation)?;
+                    let decreased_remaining = decrease_remaining_for_campaign(&redis, campaign.id, amount_to_decr).await.ok_or(Error::FailedUpdate("Could't decrease remaining".to_string()))?;
+                    // If it goes below 0 it will still return 0
+                    if decreased_remaining.eq(&UnifiedNum::from_u64(0)) {
+                        return Err(Error::NewBudget("No budget remaining after decreasing".to_string()));
```
elpiel: I have to think about this error. If we have updated redis already, this will fail and it wont update the Postgres record. We must be careful here or this could cause issues down the line

sentry/src/routes/campaign.rs
```diff
@@ -1,56 +1,286 @@
...
+
+        let latest_spendable = fetch_spendable(pool.clone(), &campaign.creator, &campaign.channel.id()).await?;
+
+        // Gets the latest Spendable for this (spender, channelId) pair
+        let total_deposited = latest_spendable.deposit.total;
+
+        let total_remaining = total_deposited.checked_sub(&accounting_spent).ok_or(Error::Calculation)?;
+        let campaigns_for_channel = get_campaigns_by_channel(&pool, &campaign.channel.id()).await?;
+        // campaign.budget should 100% exist, therefore it should be safe to just unwrap?
+        let current_campaign_budget = modified_campaign.budget.ok_or(campaign.budget).unwrap();
```
elpiel: it's better to use `unwrap_or()` in this case. It will use the `campaign.budget` and will directly return to you `UnifiedNum` instead of `Option<UnifiedNum>`

- [ ] Use `unwrap_or()`

sentry/src/routes/campaign.rs
```diff
@@ -1,56 +1,286 @@
...
+        let campaigns_remaining_sum = get_campaigns_remaining_sum(&redis, &campaigns_for_channel, campaign.id, &current_campaign_budget).await.map_err(|_| Error::Calculation)?;
```
elpiel: I'm reminding what Ivo has stated in Discord:
> also on another note, I found something that may be confusing
> > make sure sum(campaigns.map(getRemaining)) <= totalDeposited - totalSpent >
> Is an abstract constraint, not a procedure.
> 
> Meaning that it has to hold true after we mutate/set the remaining budget of the modified campaign, but we should do nothing if it doesn't (won't hold true)
> 
> To give you a better example, the check should be something like:
>
> sum(otherCampaigns.map(getRemaining)) + remainingValueForMutatedCampaign <= totalDeposited - totalSpent and if that's true, we commit remainingValueForMutatedCampaign to the DB by making adjustments to the Redis entry and the campaign entry (totalBudget)

- [x] First make this check and then continue with Mutating Campaign (redis & postgres) if the check holds true

sentry/src/routes/campaign.rs
simzzz: Can you clarify? Wouldn't the `SETNX` return `None` which would therefore return `Error` if it already exists?

elpiel: It returns:
- 1 if the key was set (true)
- 0 if the key was not set (false)

https://redis.io/commands/setnx

If it exists we should try to insert it in Postgres as it should be already there if it's in Redis. 

This shouldn't even happen as we generate a new random CampaignId when we create new campaigns (CreateCampaign) but it's good to handle this unlikely case

elpiel (approved):

PR #412, opened 2021-06-18
AIP#61 AdEx V5: Targeting rules use UnifiedNum

elpiel:


adview-manager/src/lib.rs
```diff
@@ -234,7 +234,7 @@ pub fn get_unit_html_with_events(
             .map(|validator| {
                 let fetch_url = format!(
                     "{}/channel/{}/events?pubAddr={}",
```
elpiel, line 236: This will be fixed when the `events` route for a `Campaign` is updated, see tasks for #382

primitives/src/targeting/eval.rs
```diff
@@ -196,8 +171,8 @@ impl TryFrom<SerdeValue> for Value {
         match serde_value {
             SerdeValue::Bool(bool) => Ok(Self::Bool(bool)),
             SerdeValue::Number(number) => Ok(Self::Number(number)),
-            // It's impossible to have a BigNumber literal in the rules, since they're JSON based (conform to serde_json::value::Value)
-            // However it is possible to obtain a BigNumber by invoking the Function::Bn
+            // It's impossible to have a UnifiedNum literal in the rules, since they're JSON based (conform to serde_json::value::Value)
+            // However it is possible to obtain a UnifiedNum by invoking the Function::Bn
```
simzzz, line 175: Can you still obtain UnifiedNum with `Function::Bn`?

elpiel: Yes, the `UnifiedNum` is deserialized the same way as the `BigNum` in JSON - a string.

The name of the Function::Bn will stay the same as it's more relatable than something like `un`.

simzzz (approved):

samparsky (approved):

Issue #398, opened 2021-04-19
AIP #61 v5: NewState and Accounting


elpiel:
### Sentry 

`NewState` and `Accounting` should contain 2 balances maps (for earners and spenders) and we need to be able to distinguish between them:
- balances - earners map
- balancesSpenders - spenders map

**IMPORTANT:** Although there is division in 2 maps of the balances, the `state_root` should contain both (i.e. the should be merged into a single tree)

balancesSpenders - in order to avoid duplication when merging the two maps, `spenderLeaf`s should be hashed by `hash("spender", spender, totalSpent)` (notice the string slice included when hashing to create a different hash than the earners map)

`Accounting`:
 - [x] Used to create `NewState` & `ApproveState` #428 
 - [x] Filled by the flat `sentry::db::Accounting` for a given `Channel` #439 
 - [x] MerkleTree should use actual precision of the token (e.g. DAI - 18, USDT - 6, etc.) #426

PR #413, opened 2021-06-30
Issue 382 campaign insert events

elpiel:
Part of #382 with changes in regards to #398 

### Done:

- Campaign changes
- Insert Campaign events
- Apply limits (`access.rs`)
- Sentry (with all balances) Accounting
- Sentry Database Accounting - flatten representation of the balances


sentry/src/db/accounting.rs
```diff
@@ -73,40 +177,430 @@ mod test {
...
+                spender,
+                Side::Spender,
+                update_amount,
+            )
+            .await
+            .expect("Should insert");
+            assert_eq!(spender, updated.address);
+            assert_eq!(Side::Spender, updated.side);
+            assert_eq!(
+                UnifiedNum::from(300_000_000),
```
simzzz: I think it's better to use existing variables instead of hardcoded values in the asserts (ex. `amount + updated_amount`), easier to modify and clearer when reading.

elpiel: Sounds good, will change them!
Usually I avoid using calculations, since they can add additional complexity and bugs in the tests can slip through.

simzzz (approved):

PR #416, opened 2021-07-21
GET total_deposited + spender_leaf route completed + tests

### LINKED ISSUE #391: AIP #61 v5: GET /channel/:id/spender/:addr

See the AIP#61 for up-to-date information.

### Tasks

- [x] `returns { totalDeposited, spenderLeaf }, where spenderLeaf will contain totalSpent and merkle proof` (from AIP#61)
  - `Accounting` & `NewState` include spender Balances, from which the `spenderLeaf` will be taken.
  - `spenderLeaf: (spender, balance)` however the request already contains the spender address, this is why we only need `totalSpent`
  - `merkle proof` of the **latest** `ApproveState`, this is so that we can validate that `spenderLeaf(hash("spender", spender, totalSpent))` is present in `latestApprovedState.stateRoot` (todo: additional clarifications needed)
 
- [x] `GET will return total spendable amount (from Spendable in the DB) and spender leaf` (from AIP#61)
- [x] Update NewState & ApproveState with the new `Balances::<CheckedState>` once merged #422 #423
- [x] ~Generate MerkleProof in the `SpenderLeaf`~ https://github.com/AdExNetwork/adex-validator-stack-rust/issues/377#issuecomment-933280214

simzzz:
Questions :

- Should we add a different Error struct for the `create_spendable_document()` helper function like we did for the campaign routes?
- What will happen if an error is thrown after the new Spendable is inserted in DB (ex. during total_spent calculations) and should we worry about it?
- `adapter.get_deposit()` returns a deposit that uses BigNum instead of the UnifiedNum we need for spendable. How should we deal with that?
- `NewState`uses `BalancesMap` which uses `BigNum` instead of `UnifiedMap` which uses `UnifiedNum`. How should we deal with that?


elpiel:
Q: > adapter.get_deposit() returns a deposit that uses BigNum instead of the UnifiedNum we need for spendable. How should we deal with that?
A: Using the Deposit asset of the channel we should convert the BigNum to UnifiedNum based on the precision of the token.

Q: > NewState uses BalancesMap which uses BigNum instead of UnifiedMap which uses UnifiedNum. How should we deal with that?
A: It should use `UnifiedNum`. It just hasn't been changed yet.

I will answer the other 2 questions once I review the PR.

samparsky (dismissed):

primitives/src/sentry.rs
```diff
@@ -197,6 +197,19 @@ pub struct SuccessResponse {
     pub success: bool,
 }
 
+#[derive(Serialize, Deserialize, Debug)]
+pub struct SpenderResponse {
+    pub total_deposited: Option<UnifiedNum>,
+    pub spender_leaf: Option<SpenderLeaf>,
+}
+
+// TODO: Maybe there is a better place for this struct
```
elpiel: leave here for now. It's part of the sentry response so can stay here for now.

sentry/src/routes/channel.rs
```diff
@@ -228,3 +234,139 @@ pub async fn create_validator_messages<A: Adapter + 'static>(
...
+        .get::<RouteParams>()
+        .expect("request should have route params");
+
+    let channel = req
+        .extensions()
+        .get::<ChannelV5>()
+        .expect("Request should have Channel")
+        .to_owned();
+
+    let channel_id = ChannelId::from_hex(route_params.index(0))?;
```
elpiel: - [x] use `channel.id()` instead

sentry/src/routes/channel.rs
```diff
@@ -228,3 +234,139 @@ pub async fn create_validator_messages<A: Adapter + 'static>(
...
+    let new_state = latest_new_state_v5(&app.pool, &channel, &state_root).await?.ok_or(ResponseError::BadRequest("non-existing NewState".to_string()))?;
+
+    let latest_spendable = fetch_spendable(app.pool.clone(), &spender, &channel_id)
+        .await?;
+    let latest_spendable = match latest_spendable {
+        Some(spendable) => spendable,
+        None => create_spendable_document(&app.adapter, app.pool.clone(), &channel, &spender).await?,
+    };
+
+    let total_deposited = latest_spendable.deposit.total.checked_add(&latest_spendable.deposit.still_on_create2).ok_or_else(|| ResponseError::BadRequest("Total Deposited is too large".to_string()))?;
```
elpiel: `total` includes the `still_on_creat2`, no need for the addition here.

- [x] remove addition

primitives/src/sentry.rs
```diff
@@ -197,6 +197,19 @@ pub struct SuccessResponse {
     pub success: bool,
 }
 
+#[derive(Serialize, Deserialize, Debug)]
+pub struct SpenderResponse {
+    pub total_deposited: Option<UnifiedNum>,
```
elpiel: it would be better to have a non Option field.
- [x] Use `0` for total deposit if there is no deposit.

sentry/src/lib.rs
```diff
@@ -379,6 +381,23 @@ async fn channels_router<A: Adapter + 'static>(
         req = ChannelLoad.call(req, app).await?;
 
         list_channel_event_aggregates(req, app).await
+    } else if let (Some(caps), &Method::GET) = (
+        CHANNEL_SPENDER_LEAF_AND_TOTAL_DEPOSITED.captures(&path),
+        method,
+    ) {
+        req = AuthRequired.call(req, app).await?;
```
elpiel: you can combine both middlwares and execute them at the same time.

sentry/src/routes/channel.rs
```diff
@@ -228,3 +234,139 @@ pub async fn create_validator_messages<A: Adapter + 'static>(
...
+        channel: channel.clone(),
+        deposit: Deposit {
+            total: UnifiedNum::from_u64(deposit.total.to_u64().expect("Todo")),
+            still_on_create2: UnifiedNum::from_u64(deposit.still_on_create2.to_u64().expect("Todo")),
+        },
+        spender: spender.clone(),
+    };
+
+    // Insert latest spendable in DB
+    insert_spendable(pool, &spendable).await?;
```
elpiel: What we need is not `insert_spendable` in this case.

- [x] create an `update_spendable` db fn
   It would be better to make `INSERT INTO ON CONFLICT ON CONSTRAINT ***key*** DO UPDATE SET`
   
This will be useful when we are updating the Spendable.

sentry/src/routes/channel.rs
```diff
@@ -228,3 +234,139 @@ pub async fn create_validator_messages<A: Adapter + 'static>(
...
+
+// TODO: Maybe use a different error for helper functions like in campaign routes/move this somewhere else?
+async fn create_spendable_document<A: Adapter>(adapter: &A, pool: DbPool, channel: &ChannelV5, spender: &Address) -> Result<Spendable, ResponseError> {
+    let deposit = adapter.get_deposit(&channel, &spender).await?;
+    let spendable = Spendable {
+        channel: channel.clone(),
+        deposit: Deposit {
+            total: UnifiedNum::from_u64(deposit.total.to_u64().expect("Todo")),
+            still_on_create2: UnifiedNum::from_u64(deposit.still_on_create2.to_u64().expect("Todo")),
+        },
```
elpiel: - [x] you need to fetch the `Channel.token` **precision** from `Config`
    this can fail, so you need to handle this error)

sentry/src/routes/channel.rs
```diff
@@ -228,3 +234,139 @@ pub async fn create_validator_messages<A: Adapter + 'static>(
         }
     }
 }
+
+// TODO: Maybe use a different error for helper functions like in campaign routes/move this somewhere else?
+async fn create_spendable_document<A: Adapter>(adapter: &A, pool: DbPool, channel: &ChannelV5, spender: &Address) -> Result<Spendable, ResponseError> {
```
elpiel: This will be useful for the Update ( POST ) request too.

- [ ] `fn update_spendable_from_adapter`

sentry/src/routes/channel.rs
```diff
@@ -228,3 +234,139 @@ pub async fn create_validator_messages<A: Adapter + 'static>(
...
+    // Calculate total_spent
+    let spender_balance = new_state.msg.balances.get(&spender).ok_or_else(|| ResponseError::BadRequest("No balance for this spender".to_string()))?;
+
+    let spender_balance = UnifiedNum::from_u64(spender_balance.to_u64().expect("Consider replacing this in a way that uses UnifiedMap"));
+    let total_spent = total_deposited.checked_sub(&spender_balance).ok_or_else(|| ResponseError::BadRequest("Total spent is too large".to_string()))?;
+
+    let spender_leaf = SpenderLeaf {
+        total_spent,
+        // merkle_proof: [u8; 32], // TODO
+    };
```
elpiel: Please make sure to:
- [ ] Update NewState & ApproveState with the new `Balances::<CheckedState>`
- [x] `total_spent` - The only thing we need is to return the `new_state.balances.spenders.get(spender)`

- [x] If there the spender is not found in the balances - `spender_leaf` will be `None`

sentry/src/routes/channel.rs
```diff
@@ -228,3 +234,139 @@ pub async fn create_validator_messages<A: Adapter + 'static>(
...
+    let channel = req
+        .extensions()
+        .get::<ChannelV5>()
+        .expect("Request should have Channel")
+        .to_owned();
+
+    let channel_id = ChannelId::from_hex(route_params.index(0))?;
+    let spender = Address::from_str(&route_params.index(1))?;
+
+    let default_response = Response::builder()
```
elpiel: Since all values are needed for the spender_leaf (which is an Option) you can instead match both values:

- fetch approve state
- fetch new state
- fetch the new_state.balances.spenders.get(spender) = total_spent

If any of the values is not found (i.e. is `None`), return None for `spender_leaf`

elpiel (changes requested):

elpiel:
> Should we add a different Error struct for the create_spendable_document() helper function like we did for the campaign routes?
You would probably need a new Error struct. There's probably a nicer way for those errors instead of creating new structs for Error everywhere but for the time being create a new struct.
It will hold the `A::AdapterError` and `PoolError`.

> What will happen if an error is thrown after the new Spendable is inserted in DB (ex. during total_spent calculations) and should we worry about it?

As for the `total_spent` there won't be an error as I mentioned in the review. We only return the `new_state.balances.spenders.get(spender)` there

sentry/src/routes/channel.rs
simzzz: Leaving the Balances::<CheckedState> part as TODO as it's not part of the `NewState` yet and adding it as a task in the issue.

primitives/src/spender.rs
```diff
@@ -27,23 +28,19 @@ pub struct Aggregate {
...
     use super::*;
 
-    impl TryFrom<Row> for Spendable {
-        type Error = Error;
 
-        fn try_from(row: Row) -> Result<Self, Self::Error> {
-            Ok(Spendable {
-                spender: row.try_get("spender")?,
-                channel: row.try_get("channel")?,
+    impl From<Row> for Spendable {
```
elpiel, line 45: Could you please remove the `derive(Serialize, Deserialize)` from `Spendable` as we don't use it for responses, we have the `SpendableReponse` for that

elpiel (changes requested):
Could you please add `camelCase` to both structs `SpenderResponse` & `SpenderLeaf`.
It's missing now

elpiel:
- [x] Also separate the `Spender` (to spender.rs) and the make the `SpenderResponse { spender: Spender }`

- [x] For `spender/all` then we can use: `AllSpendersReponse { spenders: HashMap<Address, Spender>  }` but later we will add pagination, so more fields will be added to the struct
- [x] Spender { total_deposited, spender_leaf: SpenderLeaf } - please change `total` to `total_deposited`

**Please note that:** we have the address in the route `/spender/:address`, this is not true for the `spender/all` route, so we need include the Address (i.e. using the `HashMap<Address, Spender>`)

primitives/src/sentry.rs
```diff
@@ -197,6 +198,19 @@ pub struct SuccessResponse {
...
 
+#[derive(Serialize, Deserialize, Debug)]
+#[serde(rename_all = "camelCase")]
+pub struct SpenderResponse {
+    pub spender: Spender,
+}
+
+#[derive(Serialize, Deserialize, Debug, Clone)]
+#[serde(rename_all = "camelCase")]
+pub struct SpenderLeaf {
```
elpiel: - [x] This should also be moved to `spender.rs`

sentry/src/routes/channel.rs
```diff
@@ -228,3 +234,169 @@ pub async fn create_validator_messages<A: Adapter + 'static>(
         }
     }
 }
+
+async fn create_spendable_document<A: Adapter + 'static>(app: &Application<A>, channel: &ChannelV5, spender: &Address) -> Result<Spendable, ResponseError> {
+    let deposit = app.adapter.get_deposit(&channel, &spender).await?;
+    let token_info = app.config.token_address_whitelist.get(&channel.token).ok_or_else(|| ResponseError::BadRequest("channel has invalid token".to_string()))?; // I don't think this error can happen
```
elpiel: it is possible to happen. E.g. a token is removed from the configuration.
This can happen since people can run their own validators and use/change their own configuration.

- [x] name the error correctly: `ResponseError::FailedValidation("Unsupported Channel token")`
BadRequest is usually intended for when bad things happen, like overflow in calculation, db error, etc. We need to make this clear as it's misused right now on many places.

sentry/src/routes/channel.rs
```diff
@@ -228,3 +234,169 @@ pub async fn create_validator_messages<A: Adapter + 'static>(
...
+    };
+
+    let state_root = approve_state.msg.state_root.clone();
+
+    let new_state = match latest_new_state_v5(&app.pool, &channel, &state_root).await? {
+        Some(new_state) => new_state,
+        None => return spender_response_without_leaf(total_deposited),
+    };
+
+    // TODO: Temporary until NewState uses Balances<CheckedState>
```
elpiel: Please 
- [ ] change NewState to use Balances<CheckedState>
Preferably in a new branch for easy reviewing based of branch `aip-61`

sentry/src/routes/channel.rs
```diff
@@ -228,3 +234,169 @@ pub async fn create_validator_messages<A: Adapter + 'static>(
...
+    let token_info = app.config.token_address_whitelist.get(&channel.token).ok_or_else(|| ResponseError::BadRequest("channel has invalid token".to_string()))?;
+
+    let total_spent = match new_state.msg.balances.get(&spender) {
+        Some(amount) => UnifiedNum::from_precision(amount.clone(), token_info.precision.get()),
+        None => Some(UnifiedNum::from_u64(0)),
```
elpiel: - [x] On `None` you should **not** return SpenderLeaf (i.e. `spender_leaf: None`)

sentry/src/routes/channel.rs
```diff
@@ -228,3 +234,169 @@ pub async fn create_validator_messages<A: Adapter + 'static>(
...
+    let new_state = match latest_new_state_v5(&app.pool, &channel, &state_root).await? {
+        Some(new_state) => new_state,
+        None => return spender_response_without_leaf(total_deposited),
+    };
+
+    // TODO: Temporary until NewState uses Balances<CheckedState>
+    let token_info = app.config.token_address_whitelist.get(&channel.token).ok_or_else(|| ResponseError::BadRequest("channel has invalid token".to_string()))?;
+
+    let total_spent = match new_state.msg.balances.get(&spender) {
+        Some(amount) => UnifiedNum::from_precision(amount.clone(), token_info.precision.get()),
```
elpiel: - [ ] handle the possible Overflow (i.e. `None) here

sentry/src/routes/channel.rs
```diff
@@ -228,3 +234,169 @@ pub async fn create_validator_messages<A: Adapter + 'static>(
...
+        let config = configuration("development", None).expect("Dev config should be available");
+        let dummy_adapter = DummyAdapter::init(adapter_options, &config);
+        setup_test_migrations(database.pool.clone())
+        .await
+        .expect("Migrations should succeed");
+        let mut channel = DUMMY_CAMPAIGN.channel.clone();
+        let token_address = Address::from_str("0x509ee0d083ddf8ac028f2a56731412edd63223b9").expect("should generate");
+        let token_info = config.token_address_whitelist.get(&token_address).expect("should retrieve address");
+        let precision: u8 = token_info.precision.into();
+        channel.token = token_address;
```
elpiel: Is there a reason to use a dedicated hardcoded token address?
- [ ] Can you just use the `channel.token`?

sentry/src/routes/channel.rs
```diff
@@ -228,3 +234,169 @@ pub async fn create_validator_messages<A: Adapter + 'static>(
...
+        let deposit = Deposit {
+            total: BigNum::from(1000000000),
+            still_on_create2: BigNum::from(1000000),
+        };
+        dummy_adapter.add_deposit_call(channel.id(), ADDRESSES["creator"], deposit.clone());
+        let campaign_remaining = CampaignRemaining::new(redis.connection.clone());
+
+        let app = Application::new(
```
elpiel: there is actually a helper function to help you with this:
setup_dummy_app - it will setup DB pools and everything
- [x] Use `setup_dummy_app`

sentry/src/routes/channel.rs
```diff
@@ -228,3 +234,169 @@ pub async fn create_validator_messages<A: Adapter + 'static>(
         }
     }
 }
+
+async fn create_spendable_document<A: Adapter + 'static>(app: &Application<A>, channel: &ChannelV5, spender: &Address) -> Result<Spendable, ResponseError> {
```
elpiel: - [x] You can directly use `Address` instead of a reference.
`Address` is Copy, which means it's cheap to Copy the value (i.e. from a reference to owned value) and the compilator can do this automatically.

elpiel: Since you use only the Adapter & DB pool from the app, you can simplify the testing setup required and the function parameters requirements by passing the 2 params instead `App`
`(adapter: &A, pool: DbPool, channel: &ChannelV5, spender: Address)`

- [x] remove `App` parameter to simplify testing

sentry/src/routes/channel.rs
```diff
@@ -228,3 +234,169 @@ pub async fn create_validator_messages<A: Adapter + 'static>(
...
+
+        let total_as_unified_num = UnifiedNum::from_precision(deposit.total, precision).expect("should convert");
+        let still_on_create2_unified = UnifiedNum::from_precision(deposit.still_on_create2, precision).expect("should convert");
+        assert_eq!(new_spendable.deposit.total, total_as_unified_num);
+        assert_eq!(new_spendable.deposit.still_on_create2, still_on_create2_unified);
+        assert_eq!(new_spendable.spender, ADDRESSES["creator"]);
+
+        // Make sure spendable NOW exists
+        let spendable = fetch_spendable(database.pool.clone(), &ADDRESSES["creator"], &channel.id()).await.expect("should return a spendable");
+        assert!(spendable.is_some());
```
elpiel, line 407: - [x] Continue the test case with scenario by updating the deposit and see if it gets reflected in `Spendable` after calling the `create_spendable_document`/`update_spendable_from_adapter`

elpiel (changes requested):
Make sure that `clippy` passes the checks & run `rustfmt`

sentry/src/routes/channel.rs
simzzz: The `channel.token` in the `DUMMY_CHANNEL` is not included in the whitelist used for development. I can add it or just change it in DUMMY_CHANNEL.

sentry/src/routes/channel.rs
simzzz: Noted. I'm resolving this and I will add a new branch.

sentry/src/routes/channel.rs
elpiel: You can use the DUMMY_CAMPAIGN & the v5 Channel does. DUMMY_CHANNEL should be fazed out soon.
`DUMMY_CAMPAIGN.channel.token`

Screenshot from 2021-08-18 18-29-56

elpiel (approved):

PR #419, opened 2021-08-13
changes to precision and tests

simzzz:


elpiel:
For the failing tests: The time has come where we hit the limit of 16 DBs in Redis and there are no available left for the tests.
This is why we get `Should return Object: Backend(OutOfBound)`
Screenshot from 2021-08-16 12-05-13


I will figure out a way to fix this but if you have ideas - share them.

PR #424, opened 2021-08-27
Issue 392 get all spenders info

### LINKED ISSUE #392: AIP #61 v5: GET /channel/:id/spender/all

Related to #391 as this should return all spenders limits 

- [x] Route impl #424 
- [x] Pagination of response #441

simzzz:


sentry/src/routes/channel.rs
```diff
@@ -342,6 +336,82 @@ pub async fn get_spender_limits<A: Adapter + 'static>(
...
+        .ok_or(ResponseError::NotFound)?;
+
+    let mut all_spender_limits: HashMap<Address, Spender> = HashMap::new();
+
+    // Using for loop to avoid async closures
+    for (spender_addr, balance) in new_state.msg.balances.spenders.iter() {
+        let latest_spendable =
+            match fetch_spendable(app.pool.clone(), spender_addr, &channel.id()).await? {
+                Some(spendable) => spendable,
+                None => return Err(ResponseError::NotFound),
```
elpiel: Instead of returning NotFound it should probably skip this spender.
This usually should not happen if there is a spender in the NewState, but it's better to return the results instead of failing on the first not-found spender

sentry/src/routes/channel.rs
```diff
@@ -342,6 +336,82 @@ pub async fn get_spender_limits<A: Adapter + 'static>(
...
+        spenders: all_spender_limits,
+    };
+
+    Ok(success_response(serde_json::to_string(&res)?))
+}
+
+async fn get_corresponding_new_state(
+    pool: &DbPool,
+    channel: &ChannelV5,
+) -> Option<MessageResponse<NewState<UncheckedState>>> {
```
elpiel: This should better return Error as well. The function `latest_approve_state_v5`/`latest_new_state_v5` already returns an `Option` if the NewState/ApproveState is not found.
If we remove the error and there is an issue with the database it will be erased.

sentry/src/routes/channel.rs
```diff
@@ -342,6 +336,82 @@ pub async fn get_spender_limits<A: Adapter + 'static>(
...
+}
+
+fn get_spender_leaf_for_spender(
+    spender_balance: &UnifiedNum,
+    total_deposited: &UnifiedNum,
+) -> Option<SpenderLeaf> {
+    let total_spent = match total_deposited.checked_sub(spender_balance) {
+        Some(spent) => spent,
+        None => return None,
+    };
```
elpiel: If I remember correctly , you can use `?` on an `Option` when your function returns an `Option`, the same way as `Result`.

```suggestion
    let total_spent = match total_deposited.checked_sub(spender_balance)?;
```

sentry/src/routes/channel.rs
```diff
@@ -342,6 +336,82 @@ pub async fn get_spender_limits<A: Adapter + 'static>(
...
+
+    let state_root = approve_state.msg.state_root.clone();
+
+    let new_state = latest_new_state_v5(pool, channel, &state_root).await.ok()?;
+
+    new_state
+}
+
+fn get_spender_leaf_for_spender(
```
elpiel: This function seems quite small and cleaner to remove it

elpiel (changes requested):

sentry/src/routes/channel.rs
```diff
@@ -342,6 +336,72 @@ pub async fn get_spender_limits<A: Adapter + 'static>(
...
+        let latest_spendable =
+            match fetch_spendable(app.pool.clone(), spender_addr, &channel.id()).await? {
+                Some(spendable) => spendable,
+                None => continue, // skipping spender if not found
+            };
+
+        let total_deposited = latest_spendable.deposit.total;
+        let total_spent = total_deposited.checked_sub(balance).ok_or_else(|| {
+            ResponseError::FailedValidation("Couldn't calculate total_spent".to_string())
+        })?;
```
```suggestion
        let total_spent = total_deposited.checked_sub(balance).ok_or_default();
```

Do not throw error if there is a spending calculation underflow. This will cause the whole sentry to stale a channel

sentry/src/routes/channel.rs
```diff
@@ -342,6 +336,72 @@ pub async fn get_spender_limits<A: Adapter + 'static>(
...
+
+    let new_state = match get_corresponding_new_state(&app.pool, &channel).await? {
+        Some(new_state) => new_state,
+        None => return Err(ResponseError::NotFound),
+    };
+
+    let mut all_spender_limits: HashMap<Address, Spender> = HashMap::new();
+
+    // Using for loop to avoid async closures
+    for (spender_addr, balance) in new_state.msg.balances.spenders.iter() {
```
elpiel: What you should do is:
1. Fetch all `spendable`s
2. If there is a value in `NewState.spenders` then add `Some(SpenderLeaf)` otherwise `None`

elpiel (changes requested):

sentry/src/routes/channel.rs
```diff
@@ -342,6 +336,74 @@ pub async fn get_spender_limits<A: Adapter + 'static>(
...
+                total_spent: spendable
+                    .deposit
+                    .total
+                    .checked_sub(balance)
+                    .unwrap_or_default(),
+                // merkle_proof: [u8; 32], // TODO
+            })
+        } else {
+            None
+        };
```
elpiel: For this case we have `Option`, you don't need to go over the trouble for checking `contains_key()` & then `unwrap()`

```suggestion
        let spender_leaf = new_state.msg.balances.spenders.get(&spender).map(|balance| {
             SpenderLeaf {
                total_spent: spendable
                    .deposit
                    .total
                    .checked_sub(balance)
                    .unwrap_or_default(),
                // merkle_proof: [u8; 32], // TODO
            }
        });
```

sentry/src/routes/channel.rs
```diff
@@ -342,6 +336,74 @@ pub async fn get_spender_limits<A: Adapter + 'static>(
...
+        Some(approve_state) => approve_state,
+        None => return Err(ResponseError::NotFound),
+    };
+
+    let state_root = approve_state.msg.state_root.clone();
+
+    let new_state = latest_new_state_v5(pool, channel, &state_root).await?;
+
+    Ok(new_state)
+}
```
elpiel: we should change this function to align with the Channel cases:

If there is no `ApproveState` - then we should return `None`. Since the channel might be new, there might not be an `ApproveState` yet.

If there is an `ApproveState` then we must fetch the `NewState`, however, since the `ApproveState` approves a given `NewState`, if `NewState` is not found we should return an error. It should not be a `NotFound` but a more critical one like `BadRequest`.

sentry/src/routes/channel.rs
```diff
@@ -342,6 +336,74 @@ pub async fn get_spender_limits<A: Adapter + 'static>(
...
+
+async fn get_corresponding_new_state(
+    pool: &DbPool,
+    channel: &ChannelV5,
+) -> Result<Option<MessageResponse<NewState<UncheckedState>>>, ResponseError> {
+    let approve_state = match latest_approve_state_v5(pool, channel).await? {
+        Some(approve_state) => approve_state,
+        None => return Err(ResponseError::NotFound),
+    };
```
```suggestion
    let approve_state = match latest_approve_state_v5(pool, channel).await? {
        Some(approve_state) => approve_state,
        None => return None,
    };
```

sentry/src/routes/channel.rs
```diff
@@ -342,6 +336,74 @@ pub async fn get_spender_limits<A: Adapter + 'static>(
...
+
+    let state_root = approve_state.msg.state_root.clone();
+
+    let new_state = latest_new_state_v5(pool, channel, &state_root).await?;
+
+    Ok(new_state)
```
```suggestion
    let new_state = match latest_new_state_v5(pool, channel, &state_root).await? {
       // TODO: Since it's an approved NewState, then it's safe to make sure it's in `CheckedState`, or if it's not - log the error that the balances are not aligned with the fact it's an Approved NewState and return ResponseError::BadRequest
        Some(new_state) => Ok(Some(new_state)),
        
        None => {
            // TODO: Log the error since this should never happen and its crucial to the Channel
            return Err(ResponseError::BadRequest("Fatal error! The NewState for the last ApproveState was not found"))
        },
    };
```

sentry/src/routes/channel.rs
```diff
@@ -342,6 +336,74 @@ pub async fn get_spender_limits<A: Adapter + 'static>(
...
+    let channel = req
+        .extensions()
+        .get::<ChannelV5>()
+        .expect("Request should have Channel")
+        .to_owned();
+
+    let new_state = match get_corresponding_new_state(&app.pool, &channel).await? {
+        Some(new_state) => new_state,
+        None => return Err(ResponseError::NotFound),
+    };
```
elpiel: Since we need to handle the use case of the Channel (see bellow) we need:

```suggestion
    let new_state = get_corresponding_new_state(&app.pool, &channel).await?; // an `Option<MessageResponse<NewState<CheckedState>>>`
```

elpiel (changes requested):

primitives/src/sentry.rs
```diff
@@ -214,6 +214,12 @@ pub struct SpenderResponse {
     pub spender: Spender,
 }
 
+#[derive(Serialize, Deserialize, Debug)]
+#[serde(rename_all = "camelCase")]
+pub struct AllSpendersResponse {
+    pub spenders: HashMap<Address, Spender>,
```
elpiel, line 227: You can add the Pagination struct for now and later we will make use of it.
For now it will be 1 page only.
```suggestion
    pub spenders: HashMap<Address, Spender>,
    #[serde(transparent)]
    pub pagination: Pagination,
```

elpiel (approved):

Issue #377, opened 2021-02-22
AIP #61: AdEx v5: significantly cheaper channels


elpiel:
### Tracking issue
For the changes regarding [AIP#61: AdEx v5: significantly cheaper channels (full details here)](https://github.com/AdExNetwork/aips/issues/61)

### Change log
- [x] `Channel` & `Campaign` #382 (PRs: #383 )
- [x] `NewState` & `Accounting` #398 
- [x] Sentry
  - [x] API documentation #475 
    - [x] #443 
    - [x] #473 
    - [x] #524 
  - [x] Accounting & spenders #381 
  - [x] GET /v5/channel/:id/accounting #431  @simzzz 
  - [x] GET /v5/channel/:id/spender/:addr #391
  - [x] GET /v5/channel/:id/spender/all #392 
  - [x] POST /v5/channel/:id/spender/:addr #465 @simzzz 
    - [x] Change output to use SpenderResponse instead of just success: true #484 @simzzz 
  - [x] POST /v5/channel/:id/pay #477 @simzzz 
  Channel Payout with authentication of the spender.
  Withdrawals of advertiser funds - re-introduces the `PAY` event with a separate route
  - [x] GET /v5/channel/:id/get-leaf @simzzz (PR #544)
    from AIP#61:
    > add another helper route /get-leaf that is called like this: /get-leaf?spender=... or /get-leaf?earner=...; this will get the latest approved state (NewState/ApproveState pair), will find the given spender/earner in the balances tree, and produce a merkle proof for it; this is useful for the Platform to verify if a spender leaf really exists
    
    I've discussed this with Ivo, we will make spender/earner routes instead of using query parameters, e.g.:
    - `/get-leaf/spender/0x....`
    - `/get-left/earner/0x...`
  - [x] POST /v5/campaign/:id/close #417 (PR #457)
    (`newBudget = totalSpent`, i.e. `newBudget` = `oldBudget - remaining`)
  - [x] POST / PUT /v5/campaign (create / modify) #382
    - update spender available funds when they try to mutate campaigns
  - [x] GET `/v5/units-for-slot` #471 
  - [x] Enforce limits ( PR #413 )
  - [x] Events #382 ( PR #413 )
    - enforce per-campaign spending limits atomically - we already do this in `access`
  - [x] Distribute fees #381 ( PR #413 )
  - [x] Analytics breakdowns by 1hr #382 
  - [x] Campaign(s) remaining #415
- [x] Validator
[...]

PR #430, opened 2021-09-16
GET /v5/campaign/list

### LINKED ISSUE #429: AIP #61 v5:  GET (list) Campaigns
Part of #382 
- Separate table for channels and campaigns, one-to-many relation with `channel_id`

- Query params:
  - `validator` - Address
  - `leader` - Address:

- [x] Changes to Query params https://github.com/AdExNetwork/adex-validator-stack-rust/pull/445

simzzz:
resolves #429


primitives/src/sentry.rs
```diff
@@ -359,6 +359,7 @@ pub mod campaign {
         pub creator: Option<Address>,
         /// filters the campaigns containing a specific validator if provided
         pub validator: Option<ValidatorId>,
+        pub is_leader: Option<bool>,
```
elpiel, line 356: - [x] Please add documentation and explain the behavior:
`false` & `None` will result in no filtration, while `Some(true)` will filter Campaigns by the provided `ValidatorId` or the authenticated `Auth.uid`

sentry/src/db/campaign.rs
```diff
@@ -54,6 +59,114 @@ pub async fn fetch_campaign(
     Ok(row.as_ref().map(Campaign::from))
 }
 
+// TODO: Export this as it's the same as the one in channel list
+struct TotalCount(pub u64);
```
elpiel: Use the same struct that Channel uses and you can move it to the `mod db`.

sentry/src/db/campaign.rs
```diff
@@ -54,6 +59,114 @@ pub async fn fetch_campaign(
...
+    accepts!(VARCHAR, TEXT);
+}
+
+pub async fn list_campaigns(
+    pool: &DbPool,
+    skip: u64,
+    limit: u32,
+    creator: &Option<Address>,
+    validator: &Option<ValidatorId>,
+    is_leader: &Option<bool>,
```
elpiel: - [x] For types that implement `Copy` you can directly pass an owned `Option` (for Address, ValidatorId & bool)

sentry/src/db/campaign.rs
```diff
@@ -54,6 +59,114 @@ pub async fn fetch_campaign(
...
+    if let Some(creator) = creator {
+        where_clauses.push(format!("creator = ${}", params.len() + 1));
+        params.push(creator);
+    }
+
+    // TODO: if no validator is provided as query param, validator will be the one from Auth, is that ok?
+    if let Some(validator) = validator {
+        where_clauses.push(format!("validators @> ${}", params.len() + 1));
+        params.push(validator);
+    }
```
elpiel: The Database query functions don't need to know about Auth, this will be additional logic when we call the given function in the Route handling.
- [x] If there no ValidatorId parameter - don't add where clause; if there is - add the clause

sentry/src/db/campaign.rs
```diff
@@ -54,6 +59,114 @@ pub async fn fetch_campaign(
...
+
+    if let (Some(validator), Some(true)) = (validator, is_leader) {
+        where_clauses.push(format!("channel->'leader' = ${}", params.len() + 1)); // TODO: change after channel is in a different table
+        params.push(validator); // TODO: could be redundant?
```
elpiel: Yes, it porbablly is. I would suggest to have this check in 1 place (i.e. above) because for the `validator` & `is_leader` you will require the same parameter - ValidatorId. Only 1 `params.push()` is required, but also the same `param_num = params.len() + 1` for the validator & is_leader checks

validator_worker/src/sentry_interface.rs
```diff
@@ -355,13 +356,15 @@ pub mod campaigns {
         sentry_url: &ApiUrl,
         page: u64,
         validator: ValidatorId,
+        is_leader: bool,
```
elpiel: This parameter is not required for this call - `None` is enough.

elpiel (changes requested):
- [x] Write tests to make sure that listing Campaigns & the total count work correctly. Generate multiple Campaigns and use a lower limit for the page results for easier impl

primitives/src/sentry.rs
elpiel: There are 2 cases where this query is taken into account:
with the query parameter `?validator`
or if an `Auth.uid` is provided (i.e. there is a auth header with token)

Please include the latter part.

sentry/src/db/campaign.rs
```diff
@@ -485,4 +582,210 @@ mod test {
...
+
+    #[tokio::test]
+    async fn it_lists_campaigns_properly() {
+        let database = DATABASE_POOL.get().await.expect("Should get a DB pool");
+
+        setup_test_migrations(database.pool.clone())
+            .await
+            .expect("Migrations should succeed");
+
+        insert_multiple_campaigns_with_different_channels(&database.pool).await;
```
elpiel: this might become a problem when a test fails since Rust does not show you properly where it has failed when calling functions with assert-ion checks.

sentry/src/db/campaign.rs
```diff
@@ -58,6 +62,95 @@ pub async fn fetch_campaign(
...
+}
+
+async fn list_campaigns_total_count<'a>(
+    pool: &DbPool,
+    (where_clauses, params): (&'a [String], Vec<&'a (dyn ToSql + Sync)>),
+) -> Result<u64, PoolError> {
+    let client = pool.get().await?;
+
+    let statement = format!(
+        "SELECT COUNT(id)::varchar FROM campaigns WHERE {}",
```
elpiel: Missing Inner join for `channels`. This is why there is an Error in the build

primitives/src/util/tests/prep_db.rs
```diff
@@ -76,6 +76,13 @@ lazy_static! {
         fee_addr: None,
     };
 
+    pub static ref DUMMY_VALIDATOR_DIFFERENT_LEADER: ValidatorDesc = ValidatorDesc {
```
elpiel: Better to set it up only for the test instead of having a static in `prep_db`

sentry/src/db/campaign.rs
```diff
@@ -58,6 +62,99 @@ pub async fn fetch_campaign(
...
+    match (validator, is_leader) {
+        (Some(validator), Some(true)) => {
+            where_clauses.push(format!("channels.leader = ${}", params.len() + 1));
+            params.push(validator);
+        }
+        (Some(validator), _) => {
+            where_clauses.push(format!(
+                "(channels.leader = ${} OR channels.follower = ${})",
+                params.len() + 1,
+                params.len() + 1
```
elpiel: You are using the same prepared value (see `params.push(validator)` - only a single push to `params`) it should be enough to have 1 `params.len()`
```suggestion
                params.len() + 1,
```

elpiel (changes requested):

sentry/src/lib.rs
```diff
@@ -184,6 +183,16 @@ impl<A: Adapter + 'static> Application<A> {
...
+            ("/v5/campaign/list", &Method::GET) => {
+                let req = match AuthRequired.call(req, self).await {
+                    Ok(req) => req,
+                    Err(error) => {
+                        return map_response_error(error);
+                    }
+                };
+
+                campaign_list(req, self).await
+            }
```
elpiel: - [x] Could you please move this to the `fn campaign_router`

sentry/src/routes/campaign.rs
```diff
@@ -214,6 +215,40 @@ pub async fn create_campaign<A: Adapter>(
...
+    let mut query =
+        serde_urlencoded::from_str::<CampaignListQuery>(req.uri().query().unwrap_or(""))?;
+
+    if query.validator.is_none() {
+        let session = req
+            .extensions()
+            .get::<Auth>()
+            .expect("Auth should exist once we're here");
+        query.validator = Some(session.uid);
+    }
```
elpiel: To have none of them is also a valid case (i.e. when user is unauthenticated and there is no validator in the query).
- [ ] If no query validator is passed, use an `Option<Auth>` - if it's not set we want to exclude this filter & Auth is not required for this route, see the #377 list of routes

sentry/src/routes/campaign.rs
```diff
@@ -217,6 +218,38 @@ pub async fn create_campaign<A: Adapter>(
...
+    app: &Application<A>,
+) -> Result<Response<Body>, ResponseError> {
+    let mut query =
+        serde_urlencoded::from_str::<CampaignListQuery>(req.uri().query().unwrap_or(""))?;
+
+    query.validator = match (query.validator, req.extensions().get::<Auth>()) {
+        (Some(validator), _) => Some(validator),
+        (None, Some(session)) => Some(session.uid),
+        (None, None) => None,
+    };
```
elpiel: `Auth` should be used as `query.validator` only when `is_leader` is passed.
By default it should not filter results by validator.

elpiel (changes requested):
- [x] Fix the Comment on the test
- [x] Fix the `query.validator` & `Auth` case.

elpiel (approved):

PR #432, opened 2021-09-17
Channel v5 table

### LINKED ISSUE #409: AIP #61 v5: Sentry POST route for inserting/modifying campaign

elpiel:


sentry/src/db/campaign.rs
```diff
@@ -39,15 +38,20 @@ pub async fn insert_campaign(pool: &DbPool, campaign: &Campaign) -> Result<bool,
 }
 
 /// ```text
-/// SELECT id, channel, creator, budget, validators, title, pricing_bounds, event_submission, ad_units, targeting_rules, created, active_from, active_to FROM campaigns
-/// WHERE id = $1
+/// SELECT campaign.id, creator, budget, validators, title, pricing_bounds, event_submission, ad_units, targeting_rules, campaign.created, active_from, active_to,
```
simzzz: should be `campaigns.id` and `campaigns.created`

sentry/src/db/campaign.rs
```diff
@@ -56,15 +60,18 @@ pub async fn fetch_campaign(
 
 // TODO: We might need to use LIMIT to implement pagination
 /// ```text
-/// SELECT id, channel, creator, budget, validators, title, pricing_bounds, event_submission, ad_units, targeting_rules, created, active_from, active_to
-/// FROM campaigns WHERE channel_id = $1
+/// SELECT campaigns.id, creator, budget, validators, title, pricing_bounds, event_submission, ad_units, targeting_rules, created, active_from, active_to,
```
simzzz: should be `campaigns.created`

sentry/src/db/campaign.rs
```diff
@@ -74,14 +81,12 @@ pub async fn get_campaigns_by_channel(
 }
 
 /// ```text
-/// UPDATE campaigns SET budget = $1, validators = $2, title = $3, pricing_bounds = $4, event_submission = $5, ad_units = $6, targeting_rules = $7
-/// WHERE id = $8
-/// RETURNING id, channel, creator, budget, validators, title, pricing_bounds, event_submission, ad_units, targeting_rules, created, active_from, active_to
+/// UPDATE campaigns SET budget = $1, validators = $2, title = $3, pricing_bounds = $4, event_submission = $5, ad_units = $6, targeting_rules = $7 FROM campaigns INNER JOIN channels ON campaigns.channel_id=channels.id WHERE campaigns.id = $8  RETURNING campaigns.id, creator, budget, validators, title, pricing_bounds, event_submission, ad_units, targeting_rules, campaign.created, active_from, active_to, channels.leader, channels.follower, channels.guardian, channels.token, channels.nonce
```
simzzz: Should be `FROM channels` instead. The comment also uses `INNER JOIN ... ON` while the statement below does not.

sentry/src/db/channel.rs
```diff
@@ -12,72 +11,63 @@ pub async fn get_channel_by_id(
 ) -> Result<Option<Channel>, PoolError> {
     let client = pool.get().await?;
 
-    let select = client.prepare("SELECT id, creator, deposit_asset, deposit_amount, valid_until, targeting_rules, spec, exhausted FROM channels WHERE id = $1 LIMIT 1").await?;
+    let select = client
+        .prepare("SELECT leader, follower, guardian, token FROM channels WHERE id = $1 LIMIT 1")
```
simzzz: Shouldn't `nonce` also be selected?

simzzz (changes requested):

sentry/src/db/channel.rs
elpiel: Yes, it should. Good catch!
:heavy_check_mark: Fixed

sentry/src/db/campaign.rs
elpiel: fixed

sentry/src/db/campaign.rs
elpiel: fixed

sentry/src/db/campaign.rs
elpiel: I've updated the full query!

simzzz (approved):

elpiel:
:raised_hands: :raised_hands: :raised_hands:

PR #439, opened 2021-09-28
Issue 431 get accounting for channel

simzzz:


sentry/src/routes/channel.rs
```diff
@@ -433,4 +475,35 @@ mod test {
...
+
+        let accounting_response = get_accounting_for_channel(req, &app).await.expect("should get response");
+        assert_eq!(StatusCode::OK, accounting_response.status());
+        let json = hyper::body::to_bytes(accounting_response.into_body())
+            .await
+            .expect("Should get json");
+
+        let accounting_response: AccountingResponse<CheckedState> =
+            serde_json::from_slice(&json).expect("Should get AccouuntingResponse");
+        let sum = accounting_response.balances.sum().expect("shouldn't be None");
```
elpiel: use better wording "shouldn't Overflow"
None is returned when there is overflow of earners or spenders.

sentry/src/routes/channel.rs
```diff
@@ -433,4 +475,35 @@ mod test {
...
+        assert_eq!(sum.0, UnifiedNum::from_u64(300));
+        assert_eq!(sum.1, UnifiedNum::from_u64(300));
+
+        update_accounting(app.pool.clone(), channel.id(), ADDRESSES["tester"], Side::Spender, UnifiedNum::from_u64(200)).await.expect("should insert accounting");
+
+        let new_req = Request::builder().extension(channel).body(Body::empty()).expect("Should build Request");
+        let bad_accounting_response = get_accounting_for_channel(new_req, &app).await;
+        assert!(bad_accounting_response.is_err())
+    }
```
elpiel: I guess you're testing a case where the `balance.check()` does not pass.

Can we improve this test to be sure that it does not fail for other reasons, e.g. db error?
Also make sure to separate / label the tests somehow.
Here is an examples:
```
fn test....() {
  ....
  
  // invalid Accounting when Balances<UncheckedState>.check() fails
  // accounting.balances.earners != accounting.balances.spenders
  {
     ... setup
     assert!(true)
  }
}

File: sentry/src/routes/channel.rs [hunk 3]
```diff
@@ -433,4 +475,35 @@ mod test {
...
+    async fn get_accountings_for_channel() {
+        let app = setup_dummy_app().await;
+        let channel = DUMMY_CAMPAIGN.channel.clone();
+        insert_channel(&app.pool, channel).await.expect("should insert channel");
+        let req = Request::builder().extension(channel).body(Body::empty()).expect("Should build Request");
+
+        update_accounting(app.pool.clone(), channel.id(), ADDRESSES["publisher"], Side::Earner, UnifiedNum::from_u64(200)).await.expect("should insert accounting");
+        update_accounting(app.pool.clone(), channel.id(), ADDRESSES["publisher2"], Side::Earner, UnifiedNum::from_u64(100)).await.expect("should insert accounting");
+        update_accounting(app.pool.clone(), channel.id(), ADDRESSES["creator"], Side::Spender, UnifiedNum::from_u64(200)).await.expect("should insert accounting");
+        update_accounting(app.pool.clone(), channel.id(), ADDRESSES["tester"], Side::Spender, UnifiedNum::from_u64(100)).await.expect("should insert accounting");
```
elpiel: Use `accounting::spend_amount` to ease this test setup.

elpiel (changes requested):
Overall needs some polishing and it will be ready!

There is a question which is good to be discussed with Ivo.
I will leave it here as a task to check it before we merge:

- [x] ~Check with Ivo if we need to validate other Route permissions in addition to having an authentication header~ no need to check with Ivo, only auth is required. should be accessibile to everyone authenticated.

elpiel (approved):

PR #445, opened 2021-10-11
Changed query params in campaign list

simzzz:


primitives/src/sentry.rs
```diff
@@ -355,10 +355,16 @@ pub mod campaign {
         pub active_to_ge: DateTime<Utc>,
         pub creator: Option<Address>,
         /// filters the campaigns containing a specific validator if provided
-        pub validator: Option<ValidatorId>,
-        /// filters the campaigns where the provided validator is a leader if true
-        /// if no validator is provided, but is_leader is true, it uses Auth to obtain a validator
-        pub is_leader: Option<bool>,
+        #[serde(flatten)]
+        pub validator: Option<ValidatorParam>,
```
elpiel, line 359: - [x] Write a small test, testing out the CampaignListQuery with and without leader/validator.

elpiel (changes requested):
LGTM, however, please make sure to test out the query (de)serialization. You might need to add a `serde(rename_all = )` attribute on `ValidatorParam`

primitives/src/sentry.rs
```diff
@@ -355,11 +355,86 @@ pub mod campaign {
...
+                creator: Some(ADDRESSES["publisher"]),
+                validator: Some(ValidatorParam::Leader(IDS["leader"])),
+            };
+
+            let query_leader_json = json!({
+                "page": 0,
+                "activeTo": 1612162800,
+                "creator": "0xB7d3F81E857692d13e9D63b232A90F4A1793189E",
+                "leader": "0xce07CbB7e054514D590a0262C93070D838bFBA2e",
+            });
```
elpiel: In the futures we might use the same IDS & ADDRESSES for all environments so please use `format!("page=.....&leader={}", IDS["leader"])` when you switch to `serde_urlencoded`

elpiel (approved):

PR #448, opened 2021-10-14
Integration tests for get_all_spendables_for_channel()

simzzz:


sentry/src/db/spendable.rs
```diff
@@ -106,7 +106,7 @@ async fn list_spendable_total_count<'a>(
     let client = pool.get().await?;
 
     let statement =
-        "SELECT COUNT(spendable.id)::varchar FROM spendable INNER JOIN channels ON spendable.channel_id=channels.id WHERE spendable.channel_id = $1";
+        "SELECT COUNT(spendable)::varchar FROM spendable INNER JOIN channels ON spendable.channel_id=channels.id WHERE spendable.channel_id = $1";
```
elpiel: You can remove the Inner join - it is not required.

elpiel (approved):
Tests could been simplified by directly creating expected values for the `Pagination` & the `Vec` and `assert_eq!` directly.

They are sufficient now as well so it's not mandatory to change them, but please remove the `INNER JOIN` for the count query

PR #457, opened 2021-11-12
Issue 417 campaign close route

### LINKED ISSUE #417: AIP #61 v5 Sentry POST /v5/campaign/:id/close

simzzz:


sentry/src/routes/campaign.rs
```diff
@@ -292,6 +293,42 @@ pub async fn campaign_list<A: Adapter>(
...
 }
 
+/// Can only be called by creator
+/// sets redis remaining = 0 (`newBudget = totalSpent`, i.e. `newBudget = oldBudget - remaining`)
+pub async fn close_campaign<A: Adapter>(
+    req: Request<Body>,
+    app: &Application<A>,
+) -> Result<Response<Body>, ResponseError> {
+    let auth = req.extensions().get::<Auth>();
+    let campaign_remaining = CampaignRemaining::new(app.redis.clone());
```
elpiel: - [x] Access it via `app`

sentry/src/routes/campaign.rs
```diff
@@ -292,6 +293,42 @@ pub async fn campaign_list<A: Adapter>(
     Ok(success_response(serde_json::to_string(&list_response)?))
 }
 
+/// Can only be called by creator
+/// sets redis remaining = 0 (`newBudget = totalSpent`, i.e. `newBudget = oldBudget - remaining`)
+pub async fn close_campaign<A: Adapter>(
+    req: Request<Body>,
+    app: &Application<A>,
+) -> Result<Response<Body>, ResponseError> {
+    let auth = req.extensions().get::<Auth>();
```
elpiel: `AuthRequired` actually makes sure that `Auth` is always present.

We should check other places too, but here it's safely to add `.expect()`

sentry/src/routes/campaign.rs
```diff
@@ -292,6 +293,42 @@ pub async fn campaign_list<A: Adapter>(
...
+        .get::<Campaign>()
+        .expect("We must have a campaign in extensions")
+        .to_owned();
+
+    let is_creator = match auth {
+        Some(auth) => auth.uid.to_address() == campaign.creator,
+        None => false,
+    };
+
+    if !is_creator {
```
elpiel: once we have the `expect` on the `get::<Auth>` this could be a simple `if auth.uid .....`

I would advice, since this is the only check, to have `if ... else ` instead of `return ...`.

sentry/src/routes/campaign.rs
```diff
@@ -1249,4 +1286,68 @@ mod test {
...
+
+            close_campaign(req, &app)
+                .await
+                .expect("Should close campaign");
+
+            let remaining = campaign_remaining
+                .get_remaining_opt(campaign.id)
+                .await
+                .expect("Should get remaining from redis")
+                .expect("There should be value for the Campaign");
```
elpiel, line 1347: Another way is to have this message in the `assert!` and use `Some(0)`:

`assert_eq!(Some(0), remaining, "There should be value for the Campaign and it should be 0 (zero)"`

sentry/src/routes/campaign.rs
```diff
@@ -1249,4 +1286,68 @@ mod test {
...
+
+            assert_eq!(remaining, 0);
+        }
+
+        // Test if an error is returned when request is not sent by creator
+        {
+            let auth = Auth {
+                era: 0,
+                uid: ValidatorId::try_from("0xce07CbB7e054514D590a0262C93070D838bFBA2e")
```
elpiel: - [ ] Please use `IDS`

sentry/src/routes/campaign.rs
```diff
@@ -1249,4 +1286,68 @@ mod test {
...
+        let multiplier = 10_u64.pow(UnifiedNum::PRECISION.into());
+        let campaign = DUMMY_CAMPAIGN.clone();
+
+        let app = setup_dummy_app().await;
+
+        // Test if remaining is set to 0
+        {
+            let campaign_remaining = CampaignRemaining::new(app.redis.clone());
+            campaign_remaining
+                .set_initial(campaign.id, UnifiedNum::from_u64(1000 * multiplier))
```
elpiel: - [ ] `CampaignRemaining` is present in the `Application`

sentry/src/routes/campaign.rs
```diff
@@ -292,6 +293,42 @@ pub async fn campaign_list<A: Adapter>(
     Ok(success_response(serde_json::to_string(&list_response)?))
 }
 
+/// Can only be called by creator
+/// sets redis remaining = 0 (`newBudget = totalSpent`, i.e. `newBudget = oldBudget - remaining`)
```
elpiel: You have forgotten to add the change of the `campaign.budget` as stated in the comment.
`newBudget = oldBudget - remaining`

elpiel (changes requested):
Let's discuss if we're going to re-use `modify_campaign` in some way or `campaign/close` can have it's own logic for setting the `budget` and then setting `remaining = 0`

sentry/src/routes/campaign.rs
```diff
@@ -292,6 +293,53 @@ pub async fn campaign_list<A: Adapter>(
...
+        ))
+    } else {
+        let remaining = app
+            .campaign_remaining
+            .get_remaining_opt(campaign.id)
+            .await?
+            .map(|remaining| UnifiedNum::from(max(0, remaining).unsigned_abs()))
+            .ok_or_else(|| {
+                ResponseError::BadRequest("No remaining entry for campaign".to_string())
+            })?;
```
elpiel: - [x] Please use https://redis.io/commands/getset
This will ensure that redis will be updated first and any subsequent events will fail with "No budget left".

sentry/src/routes/campaign.rs
```diff
@@ -292,6 +293,53 @@ pub async fn campaign_list<A: Adapter>(
...
+
+        campaign.budget = campaign.budget - remaining;
```
elpiel: - [x] Could you please use `checked_sub` and throw a BadRequest if under/overflow happens

sentry/src/routes/campaign.rs
```diff
@@ -1249,4 +1299,82 @@ mod test {
...
+        insert_channel(&app.pool, campaign.channel.clone())
+            .await
+            .expect("Should insert dummy channel");
+        insert_campaign(&app.pool, &campaign)
+            .await
+            .expect("Should insert dummy campaign");
+
+        // Test if remaining is set to 0
+        {
+            let remaining = UnifiedNum::from_u64(100 * multiplier);
```
elpiel: I would suggest to use the `campaign.budget` for this value. If we change it to something `< 100` tests will randomly start failing

sentry/src/db/campaign.rs
```diff
@@ -309,6 +309,19 @@ mod campaign_remaining {
                 .query_async(&mut self.redis.clone())
                 .await
         }
+
+        /// Doesn't allow the usage of SET with a predefined amount due to a possible race condition
+        /// use increase/decrease functions instead
```
elpiel:
```suggestion
        /// Atomic `getset` [`redis`] operation
        /// Used to close a [`Campaign`] `POST /campaign/close`
```

sentry/src/routes/campaign.rs
```diff
@@ -1249,4 +1295,81 @@ mod test {
...
+            close_campaign(req, &app)
+                .await
+                .expect("Should close campaign");
+
+            let closed_campaign = fetch_campaign(app.pool.clone(), &campaign.id)
+                .await
+                .expect("Should fetch campaign")
+                .expect("Campaign should exist");
+
+            // remaining = campaign_budget therefore sold_budget - remaining = 0
```
elpiel:
```suggestion
            // remaining == campaign_budget therefore old_budget - remaining = 0
```

elpiel (approved):

PR #458, opened 2021-11-17
Additional testing for recorder + DateHour FromSql/ToSql tests

simzzz:


primitives/src/sentry.rs
```diff
@@ -922,4 +925,33 @@ mod test {
...
+    pub async fn datehour_from_to_sql() {
+        let client = POSTGRES_POOL.get().await.unwrap();
+        let sql_type = "TIMESTAMPTZ";
+
+        let actual_datehour = DateHour::<Utc>::from_ymdh(2021, 1, 1, 1);
+
+        // from SQL
+        let row_datehour: DateHour<Utc> = client
+            .query_one(
+                &*format!("SELECT '{:?}'::{}", actual_datehour, sql_type),
```
elpiel: Debug `{:?}` seems the wrong type of formatting for the SQL.

primitives/src/sentry.rs
```diff
@@ -922,4 +925,33 @@ mod test {
             serde_json::to_value(click).expect("should serialize")
         );
     }
+
+    #[tokio::test]
+    pub async fn datehour_from_to_sql() {
+        let client = POSTGRES_POOL.get().await.unwrap();
+        let sql_type = "TIMESTAMPTZ";
+
+        let actual_datehour = DateHour::<Utc>::from_ymdh(2021, 1, 1, 1);
```
elpiel: You'd need a second value to compare the value you expect with the actual value that you get from Postgres and from postgres.
You can check the same SQL (FromSql) value against what you get back as a DateTime (with no minutes and seconds) and that DateHour correctly inserts creates an Sql value (ToSql).


Start with a `DateTime` and set `ymd().hms()` so that it's equal to the DateHour (you can also assert this). `to_datetime` will be useful for this scenario.

sentry/src/analytics.rs
```diff
@@ -152,76 +152,222 @@ mod test {
...
+                click_analytics.payout_amount,
+                UnifiedNum::from_u64(2_000_000)
+            );
+            assert_eq!(click_analytics.payout_count, 2);
+
+            assert_eq!(
+                impression_analytics.payout_amount,
+                UnifiedNum::from_u64(2_000_000)
+            );
+            assert_eq!(impression_analytics.payout_count, 2);
```
elpiel: Just realized that this can cause some issues when tests are run around `XX:59:59` since they rely that the events will be recorded with the same DateHour.
Instead, you can just use 2 of the same Events in the `input_events`, e.g. 
```
            let input_events = vec![
                test_events["click_empty"].clone(),
                test_events["impression_empty"].clone(),
                test_events["impression_empty"].clone(),
            ];
```

sentry/src/analytics.rs
```diff
@@ -152,76 +152,222 @@ mod test {
...
+            let session = Session {
+                ip: Default::default(),
+                country: Some("Bulgaria".into()),
+                referrer_header: Some("http://127.0.0.1".into()),
+                os: Some("Windows".into()),
+            };
+
+            let other_session = Session {
+                ip: Default::default(),
+                country: Some("Japan".into()),
```
elpiel: Nice :D

sentry/src/analytics.rs
```diff
@@ -152,76 +152,222 @@ mod test {
...
+                .filter(|a| a.ad_unit == Some(DUMMY_IPFS[2]) && a.ad_slot == Some(DUMMY_IPFS[3]))
+                .collect();
+            assert_eq!(with_different_slot_and_unit.len(), 1);
+
+            record(
+                &database.clone(),
+                &campaign,
+                &other_session,
+                input_events.clone(),
+            )
```
elpiel: Same problem with this record. It might end up with bad results if ran around `XX:59:59`

sentry/src/analytics.rs
```diff
@@ -152,76 +152,222 @@ mod test {
...
+                .await
+                .expect("should get all analytics");
+
+            assert_eq!(analytics.len(), 5);
+
+            let with_slot_and_unit: Vec<Analytics> = analytics
+                .clone()
+                .into_iter()
+                .filter(|a| a.ad_unit == Some(DUMMY_IPFS[0]) && a.ad_slot == Some(DUMMY_IPFS[1]))
+                .collect();
```
elpiel: - It is not necessary to clone the whole Vec before filtering. - use no `clone()` with `iter()` and after this you can `.clone()` only the resulting values and collect them. (or collect the refs. and clone() to get owned)
- I think that this setup could be cleaned up a bit and instead of `filter()`-ing all the time you could instead find the corresponding Analytics once and assert_eq! an expected Analytics with the found one. This will give us more information that just checking the expected count of analytics in the results.

Test with multiple input events that will end up in the same Analytics to avoid the problem with 59:59 time.

elpiel (changes requested):

elpiel (approved):

PR #461, opened 2021-11-29
Analytics routes

simzzz:
To-Do:

- [x] Discuss if we are going to just leave query start/end fields to be milliseconds


docs/config/ganache.toml
```diff
@@ -41,7 +41,7 @@ ethereum_adapter_relayer = 'http://localhost:8888'
 
 creators_whitelist = []
 validators_whitelist = []
-
+admins = ['0xce07CbB7e054514D590a0262C93070D838bFBA2e']
```
elpiel: You can use the real ganache ID.
Find it here:
https://github.com/AmbireTech/adex-validator-stack-rust/blob/bfdc4f75df7783350bc67ce5418ad0a79c747c93/primitives/src/test_util.rs?_pjax=%23js-repo-pjax-container%3Afirst-of-type%2C%20div%5Bitemtype%3D%22http%3A%2F%2Fschema.org%2FSoftwareSourceCode%22%5D%20main%3Afirst-of-type%2C%20%5Bdata-pjax-container%5D%3Afirst-of-type#L8

primitives/src/analytics.rs
```diff
@@ -91,6 +154,34 @@ pub enum OperatingSystem {
...
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize, Display)]
+pub enum Metric {
+    Count,
+    Paid,
+}
+
+impl Timeframe {
+    pub fn get_period_in_hours(&self) -> i64 {
```
elpiel: keep it simple: `to_hours`, also consider using u32, u64

primitives/src/sentry.rs
```diff
@@ -203,6 +203,14 @@ pub struct Analytics {
     pub payout_count: u32,
 }
 
+// TODO: Verify this is the needed output
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq)]
+#[serde(rename_all = "camelCase")]
+pub struct FetchedAnalytics {
+    pub payout_amount: Option<UnifiedNum>,
+    pub payout_count: Option<u32>,
+}
```
elpiel:
```
.then(aggr => ({
			limit: appliedLimit,
			limitReached: appliedLimit === aggr.length,
			aggr
		}))
```

While the aggr is:

```
const resultProjection = {
		value: '$value',
		time: {
			$toLong: {
				$dateFromParts: {
					year: { $ifNull: ['$_id.time.year', 0] },
					month: { $ifNull: ['$_id.time.month', 1] },
					day: { $ifNull: ['$_id.time.day', 1] },
					hour: { $ifNull: ['$_id.time.hour', 0] },
					timezone: timezone || 'UTC'
				}
			}
		},
		segment: '$_id.segment',
		_id: 0
	}
```
You can read what each part does, like `toLong` in Mongo docs to see what we should return at the end:
https://docs.mongodb.com/manual/reference/operator/aggregation/toLong/

It will be even better if you double check what the JS implementation returns as a json at the end.

sentry/src/db/analytics.rs
```diff
@@ -1,230 +1,73 @@
...
     pool: &DbPool,
-    analytics_type: AnalyticsType,
-    segment_by_channel: bool,
-    channel_id: Option<&ChannelId>,
-) -> Result<Vec<AnalyticsData>, PoolError> {
+    start_date: DateHour<Utc>,
+    end_date: Option<DateHour<Utc>>,
+    query: &AnalyticsQuery,
+    auth_as_key: Option<String>,
+    auth_uid: ValidatorId,
```
elpiel: You can use an enum for this. `AuthenticateAs::Advertiser(Address)`, ``AuthenticateAs::Publisher(Address)``

sentry/src/db/analytics.rs
```diff
@@ -1,230 +1,73 @@
...
-        }
+    let (select_clause, group_clause) = match &query.segment_by {
+        Some(segment_by) => (
+            format!("{}, time, {}", &query.metric.to_string(), segment_by),
+            format!("time, {}", segment_by),
+        ),
+        None => (
+            format!("{}, time", query.metric.to_string()),
+            "time".to_string(),
+        ),
```
elpiel: Use a vec for each of the clauses. This complicated query needs more than just strings.
You can still build the 2 vec based on the segment if you want to keep the ordering of the clauses (for groups in particular):
```
let mut select = vec!["time".to_string(), query.metric.to_string()];
let mut order_by = vec!["time".to_string()];

match Some(segment_by) = segment_by {
select.push(...);
order_by.push(...);
}
```

When building select SQL:

select.join(',')
order_bu.join(',')

sentry/src/db/analytics.rs
```diff
@@ -1,230 +1,73 @@
...
+        where_clauses.push(format!("{} = {}", auth_as_key, auth_uid))
+    }
 
+    if let Some(end_date) = end_date {
+        where_clauses.push(format!("time <= {}", end_date.to_datetime()));
+    }
     where_clauses.extend(vec![
-        format!("event_type = ${}", params.len()),
-        format!("{} IS NOT NULL", metric),
+        format!("event_type = ${}", query.event_type),
```
elpiel: $1 $2 $3 is for prepared statements.
You've removed `params.len()` so the number for the parameter in the SQL query will be wrong.

elpiel (changes requested):

primitives/src/analytics.rs
```diff
@@ -76,10 +77,72 @@ pub struct AnalyticsQuery {
...
     #[serde(default = "default_metric")]
-    pub metric: String,
+    pub metric: Metric,
     #[serde(default = "default_timeframe")]
-    pub timeframe: String,
-    pub segment_by_channel: Option<String>,
+    pub timeframe: Timeframe,
+    pub segment_by: Option<String>,
+    pub start: Option<DateTime<Utc>>,
+    pub end: Option<DateTime<Utc>>,
```
elpiel: We should probably deserialize directly into DateHour.

Also, the `Joi` package usually supports both date (String) & milliseconds (number) but we have to double check if we are going to just leave it to milliseconds (probably with @Ivshti ).

We use milliseconds almost everywhere in the JSON representaiton.

for more details check:
https://joi.dev/api/?v=17.5.0#date

primitives/src/analytics.rs
```diff
@@ -76,10 +77,72 @@ pub struct AnalyticsQuery {
...
+    // #[serde(default = "default_timezone")]
+    // pub timezone: String,
+    pub campaign_id: Option<String>,
```
elpiel: this is not a string.

primitives/src/analytics.rs
```diff
@@ -76,10 +77,72 @@ pub struct AnalyticsQuery {
...
+    pub ad_unit: Option<String>,
+    pub ad_slot: Option<String>,
```
elpiel: `IPFS`

primitives/src/analytics.rs
```diff
@@ -76,10 +77,72 @@ pub struct AnalyticsQuery {
...
+    pub ad_slot_type: Option<String>,
+    pub advertiser: Option<String>,
+    pub publisher: Option<String>,
```
elpiel: `Address`

primitives/src/analytics.rs
```diff
@@ -76,10 +77,72 @@ pub struct AnalyticsQuery {
...
+    pub hostname: Option<String>,
+    pub country: Option<String>,
+    pub os_name: Option<String>,
```
elpiel: `OperatingSystem`

elpiel (changes requested):

primitives/src/analytics.rs
simzzz: it's always positive and it's used in `Duration::hours()` which takes i64

elpiel:
TODOs:
- [x] Check (aka test) impls for Ord, PartialOrd, PartialEq for DateHour
- [x] More tests for `query::Time`
- [ ] Deserialize DateHour from either timestamp or ISO format
- [ ] Caching analytics (if needed)

simzzz:
LGTM, I think the changes reduce the complexity of the logic we had in some places.

elpiel (approved):

elpiel:
> TODOs:
> 
>     * [x]  Check (aka test) impls for Ord, PartialOrd, PartialEq for DateHour
> 
>     * [x]  More tests for `query::Time`
> 
>     * [ ]  Deserialize DateHour from either timestamp or ISO format
> 
>     * [ ]  Caching analytics (if needed)

I'm moving the last 2 TODOs to the #382 issue, these can be implemented on later stage.

PR #465, opened 2022-01-10
POST /v5/channel/:id/spender/:addr + tests

simzzz:


sentry/src/routes/channel.rs
```diff
@@ -596,4 +629,105 @@ mod test {
...
+            .await
+            .expect("should insert channel");
+
+        let build_request = |channel: Channel| {
+            Request::builder()
+                .extension(channel)
+                .body(Body::empty())
+                .expect("Should build Request")
+        };
+        let build_request_with_params = |channel: Channel| {
```
elpiel: Since `build_request` & `build_request_with_params` are used to build 2 requests for 2 different routes, I would suggest we rename them to something less generic as a name.

`add_spender_request`
`get_accounting_request`

sentry/src/routes/channel.rs
```diff
@@ -596,4 +629,105 @@ mod test {
...
+            let param = RouteParams(vec![
+                channel.id().to_string(),
+                ADDRESSES["creator"].to_string(),
```
elpiel: Refrain from using `ADDRESSES` and instead use the static in the module:
`primitives::test_util::{ADVERTISER, CREATOR, PUBLISHER, PUBLISHER_2, ...}`

sentry/src/routes/channel.rs
```diff
@@ -596,4 +629,105 @@ mod test {
...
+        // Balances shouldn't change
+        assert_eq!(
+            accounting_response
+                .balances
+                .spenders
+                .get(&ADDRESSES["creator"]),
+            balances.spenders.get(&ADDRESSES["creator"]),
+        );
+
+        // TODO: Test if updated field has changed and/or validator worker test regarding this route
```
elpiel: Still one todo left. I don't know what it is referring to, though.

elpiel (changes requested):

elpiel (approved):

PR #466, opened 2022-01-11
Adview Manager V5 changes

simzzz:
For now, it's just the changed route for `get_unit_html_with_events()` and the updated Event struct but I will also use this PR if there are more changes needed.


adview-manager/src/lib.rs
```diff
@@ -231,7 +224,7 @@ pub fn get_unit_html_with_events(
             .iter()
             .map(|validator| {
                 let fetch_url = format!(
-                    "{}/channel/{}/events?pubAddr={}",
+                    "{}/campaign/{}/events?pubAddr={}",
```
elpiel, line 230: So what should we do with the `pubAddr`?

adview-manager/src/lib.rs
```diff
@@ -211,18 +200,22 @@ pub fn get_unit_html_with_events(
...
+                referrer: Some("document.referrer".to_string()),
+            },
+            _ => Event::Impression {
+                publisher: options.publisher_addr,
+                ad_unit: Some(ad_unit.id),
+                ad_slot: Some(options.market_slot),
+                referrer: Some("document.referrer".to_string()),
+            }
+        };
+        let body = serde_json::to_string(&vec![event])
```
elpiel: As far as I remember the POST .../event takes a body of `{events: [....]}` and not `[....]`. Please double check that and make the necessary adjustments

elpiel (changes requested):

adview-manager/src/lib.rs
simzzz: It's not used in the validator but it's used for analytics/cloudflare

elpiel (approved):

PR #469, opened 2022-01-25
Validator worker test coverage

simzzz:
TODO's:

- [x] Fix the last test case which is failing
- [x] ~~Maybe move `test_leader_and_follower_tick()` into a new file~~ Call function through run_full_test() instead
- [x] Compare the PropagationResult in the last test case


sentry/Cargo.toml
```diff
@@ -28,13 +28,21 @@ hex = "0.4"
...
 regex = "1"
 # Database
 redis = { version = "0.20", features = ["aio", "tokio-comp"] }
 deadpool = "0.8.0"
 deadpool-postgres = "0.9.0"
-tokio-postgres = { version = "0.7.0", features = ["with-chrono-0_4", "with-serde_json-1"] }
+tokio-postgres = { version = "0.7.0", features = [
+  "with-chrono-0_4",
+  "with-serde_json-1",
+] }
```
elpiel: I have the feeling that this is cause by an older version of `cargo-make`. I might remove the formatting of the toml files, but can you please revert these changes?

test_harness/src/lib.rs
```diff
@@ -884,6 +900,396 @@ mod tests {
             .send()
             .await?)
     }
+
+    // Potential TODO's: Merge both test cases into one and find a way to retrieve heartbeats
```
elpiel: You can retrieve the heartbeats using `get_our_latest_msg` after running a single tick.

test_harness/src/lib.rs
```diff
@@ -884,6 +900,396 @@ mod tests {
...
+                .unwrap();
+
+            let newstate = NewState::<CheckedState>::try_from(newstate).expect("Should convert");
+            let newstate_follower =
+                NewState::<CheckedState>::try_from(newstate_follower).expect("Should convert");
+
+            assert_eq!(
+                newstate.state_root, newstate_follower.state_root,
+                "Stateroots of the new messages match"
+            );
```
elpiel: It does make some sense to test if the propagated NewState is returned, however, it would be better to submit new events and expect a NewState returned with the correct payouts instead.

test_harness/src/lib.rs
```diff
@@ -884,6 +900,396 @@ mod tests {
...
+            assert_eq!(
+                newstate
+                    .balances
+                    .earners
+                    .get(&*PUBLISHER)
+                    .expect("Should retrieve earner balance"),
+                &UnifiedNum::from_u64(18000),
+                "Publisher earner balances have been updated"
+            );
+        }
```
elpiel: Can't you just created Expected balances and `assert_eq!` them with the actual ones?

elpiel (changes requested):
I've forgot about this PR, sorry about that, next time you can ping me ;)

Instead of faking the accountings we have to work with real-life scenarios, the same way that the real system will work.
So create events for those campaigns and expect a certain message & balances at the end

elpiel:
see #486

PR #472, opened 2022-02-15
Issue #336 Multi-chain support

### LINKED ISSUE #336: Multi-chain support (EVM chains)
See https://github.com/AdExNetwork/adex-validator/issues/324 for more details
Chain IDs: https://chainid.network/

When working with Channels we need to try and map the known addresses of stablecoins to the passed `Channel.token` and "guess" the chain on which this channel is located (ethereum, polygon, etc).

## ✅ Configuration

Configuration should allow for setting up allowed tokens (and chains). So configuration for allowed tokens and chains is explicit and we have to specify the **chain id** and **address**.
Each Chain should also have the addresses of the `OUTPACE` & `Sweeper` contracts

## ✅ Adapter

1. RPC

The information about the `Chain` will be then used to initialize the `Adapter` with the correct rpc url for this chain.

2. Authentication token

Chain will be used to generate to Authentication token as well, since we need this information, when we're calling the `Adapter::session_from_token`. This is the case, because of #452 which uses the Identity contract (on a given chain) if the payload contains an identity address

## Sentry

All calls should be filtered out based on the Authentication token Chain, otherwise the identity as which we've authenticated as in the given chain, won't be valid for other chain's records (like channels, campaigns, analytics, etc)

elpiel:
This is the initial implementation of Multi-chain support.
Partially resolves #336

- ChainOf - can hold information about the Chain & Token of a given value, i.e. Campaign or Channel
- channel_context / campaign_context usually hold this value in the `context` filed.
- **NOTE**: Naming could be changed
- 

Still needs to be done:
- [ ] Database - decide where and how the ChainId is needed in the database.
- [ ] Change all SQL Queries to allow the use of ChainId
- [ ] Ganache `#1` (chain id: 1) setup with snapshot, contracts, etc. ( we use the default `#1337` right now)
- [ ] test_harness - use at lease 2 chains for tests


validator_worker/src/sentry_interface.rs
```diff
@@ -66,70 +88,106 @@ impl<C: Unlocked, P: Clone> Clone for SentryApi<C, P> {
...
+        let sentry_api = SentryApi::new(adapter, logger, config, sentry_url)?;
 
-        let sentry_api = SentryApi::new(adapter, logger, config, whoami)?;
-
-        Ok(sentry_api.with_propagate(propagate_to))
+        sentry_api.with_propagate(propagate_to)
     }
 
-    /// If the _Who am I_ Validator is not found in `propagate_to` it will add it.
+    /// If the _Who am I_ Validator is not found in `propagate_to` it will added.
```
simzzz: Should be `be added`

adapter/src/ethereum/client.rs
```diff
@@ -723,19 +722,51 @@ mod test {
 
         let spender = *CREATOR;
 
+        // todo: Clean up
```
simzzz: There's one todo left. Is it for future reference?

validator_worker/src/sentry_interface.rs
elpiel: fixed.

adapter/src/ethereum/client.rs
elpiel: Yes, could be improved later on, it's only for the test.
I've removed the todo

PR #477, opened 2022-03-01
POST /v5/channel/:id/pay

### LINKED ISSUE #377: AIP #61: AdEx v5: significantly cheaper channels

### LINKED ISSUE #382: AIP #61 v5: Channel and Campaign

simzzz:
TODO:
- [x] Verify logic
- [x] Testing
- [x] Documentation

Questions: 
How do we handle an empty body? Right now the validator will return OK without modifying anything.


sentry/src/routes/channel.rs
```diff
@@ -392,6 +398,83 @@ pub async fn get_accounting_for_channel<C: Locked + 'static>(
     Ok(success_response(serde_json::to_string(&res)?))
 }
 
+pub async fn channel_payout<C: Locked + 'static>(
+    req: Request<Body>,
+    app: &Application<C>,
+) -> Result<Response<Body>, ResponseError> {
+    let channel = req
+        .extensions()
+        .get::<Channel>()
```
elpiel: You should be using `ChainOf`

sentry/src/routes/channel.rs
```diff
@@ -392,6 +398,83 @@ pub async fn get_accounting_for_channel<C: Locked + 'static>(
...
+        .iter()
+        .sum::<Option<UnifiedNum>>()
+        .ok_or_else(|| ResponseError::BadRequest(
+            "Couldn't sum remaining amount for all campaigns".to_string(),
+        ))?;
+
+    // A campaign is closed when its remaining == 0
+    // therefore for all campaigns for a channel to be closed their total remaining sum should be 0
+    if campaigns_remaining_sum.ge(&UnifiedNum::from_u64(0)) {
+        return Err(ResponseError::BadRequest(
```
elpiel: We need Validation response error here. Closed & no budget left are the 2 cases, so maybe add that to the message:
"All campaigns should be closed or have no budget left"

sentry/src/routes/channel.rs
```diff
@@ -392,6 +398,83 @@ pub async fn get_accounting_for_channel<C: Locked + 'static>(
...
+                if accounting.amount.lt(&amount) {
+                    return Err(ResponseError::BadRequest(format!(
+                        "Spender amount for {} is less than the requested withdrawal amount",
+                        address
+                    )));
+                }
+                // If so, we add it to the earner balances for the same address
+                unchecked_balances
+                    .earners
+                    .insert(accounting.address, accounting.amount);
```
elpiel: using `earners.insert()` will never pass the `check()` call!

You should be using `spend()` and it's better to directly use the `CheckedState` instead of unchecked.

sentry/src/routes/channel.rs
```diff
@@ -392,6 +398,83 @@ pub async fn get_accounting_for_channel<C: Locked + 'static>(
...
+
+    let balances = match unchecked_balances.check() {
+        Ok(balances) => balances,
+        Err(error) => {
+            error!(&app.logger, "{}", &error; "module" => "channel_payout");
+            return Err(ResponseError::FailedValidation(
+                "Earners sum is not equal to spenders sum for channel".to_string(),
+            ));
+        }
+    };
```
elpiel: This can go away.

sentry/src/routes/channel.rs
```diff
@@ -392,6 +398,83 @@ pub async fn get_accounting_for_channel<C: Locked + 'static>(
...
+) -> Result<Response<Body>, ResponseError> {
+    let channel = req
+        .extensions()
+        .get::<Channel>()
+        .expect("Request should have Channel")
+        .to_owned();
+
+    let body = hyper::body::to_bytes(req.into_body()).await?;
+
+    let body = serde_json::from_slice::<ChannelPayBody>(&body)
```
elpiel: can you use more descriptive name like `to_pay`, `payouts`, etc?

sentry/src/routes/channel.rs
```diff
@@ -392,6 +398,83 @@ pub async fn get_accounting_for_channel<C: Locked + 'static>(
...
+            get_accounting(app.pool.clone(), channel.id(), address, Side::Spender).await?;
+        match accounting {
+            Some(accounting) => {
+                // We need to check if the amount this spender has is available for withdrawal
+                if accounting.amount.lt(&amount) {
+                    return Err(ResponseError::BadRequest(format!(
+                        "Spender amount for {} is less than the requested withdrawal amount",
+                        address
+                    )));
+                }
```
elpiel: (To be confirmed with Ivo) Most probably the logic should be as follow:

Max payout for Auth Address and Channel = Deposit - accounting.spent + accounting.earned

Then check if `max_payout <= request_payouts.sum()`

sentry/src/routes/channel.rs
```diff
@@ -392,6 +398,83 @@ pub async fn get_accounting_for_channel<C: Locked + 'static>(
...
+            error!(&app.logger, "{}", &error; "module" => "channel_payout");
+            return Err(ResponseError::FailedValidation(
+                "Earners sum is not equal to spenders sum for channel".to_string(),
+            ));
+        }
+    };
+
+    let res = spend_amount(app.pool.clone(), channel.id(), balances).await?;
+
+    Ok(success_response(serde_json::to_string(&res)?))
```
elpiel: There might not be much benefit in showing the current accountings for Earner & Spenders involved in the `spend_amount` call.
Also, the returned accounting sums is not guaranteed to be equal.

primitives/src/sentry.rs
```diff
@@ -576,6 +576,11 @@ pub struct AllSpendersQuery {
     pub page: u64,
 }
 
+#[derive(Debug, Serialize, Deserialize)]
+pub struct ChannelPayBody {
```
elpiel: `ChannelPayRequest` should be enough, what do you think?

It can be also `ChannelPay`, since we only have 1 struct with suffix `Request`, but as long as we are consistent it should be fine.

primitives/src/sentry.rs
```diff
@@ -576,6 +576,11 @@ pub struct AllSpendersQuery {
...
+    pub payouts: HashMap<Address, UnifiedNum>,
```
elpiel: please use UnifiedMap.

elpiel (changes requested):
The main logic should be adjusted as follows:

`max payout = Deposit - accounting spent + accounting earned`

`max payout <= request.payouts.sum()`

sentry/src/routes/channel.rs
simzzz: What would be a good response here? I'm thinking either `success: true` or to have it be the same as the request body.

elpiel: Let's go with `SuccessResponse`. 
Later on if needed - we can change it.

sentry/src/db/accounting.rs
```diff
@@ -27,7 +28,7 @@ impl From<tokio_postgres::Error> for Error {
     }
 }
 
-#[derive(Debug, Clone, PartialEq, Eq)]
+#[derive(Debug, Clone, PartialEq, Eq, Serialize)]
```
elpiel: you can remove the Serialize

sentry/src/db/accounting.rs
```diff
@@ -50,7 +51,7 @@ impl From<&Row> for Accounting {
     }
 }
 
-#[derive(Debug, Clone, Copy, ToSql, FromSql, PartialEq, Eq)]
+#[derive(Debug, Clone, Copy, ToSql, FromSql, PartialEq, Eq, Serialize)]
```
elpiel: You can remove the Serialize

sentry/src/routes/channel.rs
```diff
@@ -392,6 +398,95 @@ pub async fn get_accounting_for_channel<C: Locked + 'static>(
...
+    let channel_campaigns = fetch_campaign_ids_for_channel(
+        &app.pool,
+        channel_context.context.id(),
+        app.config.campaigns_find_limit,
+    )
+    .await?;
+
+    let campaigns_remaining_sum = app
+        .campaign_remaining
+        .get_multiple(channel_campaigns.as_slice())
```
elpiel: FYI: You don't usually need `as_slice()`, rust will automagically do this when you use `&` for a `Vec<T>` when you need `&[T]`

Do check out the `Deref` trait

sentry/src/routes/channel.rs
```diff
@@ -392,6 +398,95 @@ pub async fn get_accounting_for_channel<C: Locked + 'static>(
...
+        .await?
+        .iter()
+        .sum::<Option<UnifiedNum>>()
+        .ok_or_else(|| {
+            ResponseError::BadRequest("Couldn't sum remaining amount for all campaigns".to_string())
+        })?;
+
+    // A campaign is closed when its remaining == 0
+    // therefore for all campaigns for a channel to be closed their total remaining sum should be 0
+    if campaigns_remaining_sum.ge(&UnifiedNum::from_u64(0)) {
```
elpiel: please just use the `>`, this check is wrong as you use Greater or Equal. It can't be Equal to 0, because all request will fail.

sentry/src/routes/channel.rs
```diff
@@ -392,6 +398,95 @@ pub async fn get_accounting_for_channel<C: Locked + 'static>(
...
+            Side::Spender,
+        )
+        .await?
+        .map(|accounting_spent| accounting_spent.amount)
+        .unwrap_or_default();
+        let accounting_earned = get_accounting(
+            app.pool.clone(),
+            channel_context.context.id(),
+            address,
+            Side::Spender,
```
elpiel: Bug! It should be `Earner`

sentry/src/routes/channel.rs
```diff
@@ -392,6 +398,95 @@ pub async fn get_accounting_for_channel<C: Locked + 'static>(
...
+                .map_err(|err| ResponseError::BadRequest(err.to_string()))?;
+        let total_deposited = latest_spendable.deposit.total;
+
+        let available_for_return = total_deposited
+            .checked_sub(&accounting_spent)
+            .ok_or_else(|| ResponseError::FailedValidation("No more budget remaining".to_string()))?
+            .checked_add(&accounting_earned)
+            .ok_or_else(|| {
+                ResponseError::FailedValidation("No more budget remaining".to_string())
+            })?;
```
elpiel: We should be smarter here.
If `accounting_spend > total_deposit` but with the addition of `accounting_earned` it is valid, then we should still allow this computation to pass.

Now if `accounting_spent > total_deposited` the request will fail, even if we have enough from `accounting_earned`

sentry/src/routes/channel.rs
```diff
@@ -392,6 +398,95 @@ pub async fn get_accounting_for_channel<C: Locked + 'static>(
...
+        // We need to check if the amount this spender has is available for withdrawal
+        if available_for_return.lt(&amount_to_return) {
```
elpiel: Please use `<`

sentry/src/routes/channel.rs
```diff
@@ -392,6 +398,95 @@ pub async fn get_accounting_for_channel<C: Locked + 'static>(
...
+            return Err(ResponseError::BadRequest(format!(
+                "Spender amount for {} is less than the requested withdrawal amount",
+                address
+            )));
+        }
+        // If so, we add it to the earner balances for the same address
+        balances.spend(address, address, available_for_return)?;
```
elpiel: **No**, you should use the `auth.id` as he's spending this amount.

sentry/src/routes/channel.rs
```diff
@@ -392,6 +398,95 @@ pub async fn get_accounting_for_channel<C: Locked + 'static>(
...
+    // A campaign is closed when its remaining == 0
+    // therefore for all campaigns for a channel to be closed their total remaining sum should be 0
+    if campaigns_remaining_sum.ge(&UnifiedNum::from_u64(0)) {
+        return Err(ResponseError::FailedValidation(
+            "All campaigns should be closed or have no budget left".to_string(),
+        ));
+    }
+
+    let mut balances: Balances<CheckedState> = Balances::new();
+    for (address, amount_to_return) in to_pay.payouts {
```
elpiel: You must perform the queries for accounting_spent, accounting_earner, etc. for `Auth.uid` **not** the `payouts` Address!

elpiel (changes requested):
The `for` loop logic is messed up.

sentry/src/routes.rs
```diff
@@ -76,9 +76,16 @@
 //!
 //! Response: [`LastApprovedResponse`][primitives::sentry::LastApprovedResponse]
 //!
-//! - `POST /v5/channel/:id/pay` (auth required)
+//! - [`POST /v5/channel/:id/pay`](channel::channel_payout)
```
elpiel:
```suggestion
//! - [`POST /v5/channel/:id/pay`](channel::channel_payout) (auth required)
```

sentry/src/routes.rs
```diff
@@ -76,9 +76,16 @@
...
 //!
-//! TODO: implement and document as part of issue #382
+//! This route handles withdrawals of advertiser funds for a spender. Has authentication of the spender.
```
elpiel: it can handle all withdraws, not only advertiser.
Can you rephrase the sentence?

I think we can even say `for the authenticated spender` and drop the second sentence.

sentry/src/routes.rs
```diff
@@ -76,9 +76,16 @@
...
+//! It needs to ensure all campaigns are closed. It accepts a JSON body in the request which contains
+//! all of the earners and updates their balances accordingly. Used when an advertiser/spender wants
+//! to get their remaining funds back.
+//!
+//! Request JSON body: [`ChannelPayRequest`](primitives::sentry::ChannelPayRequest)
+//!
+//! Response: [`SuccessResponse`](primitives::sentry::SuccessResponse)
 //!
 //! Channel Payout with authentication of the spender
```
elpiel: I do think we can have leave this short summary of the route in the beginning of the documentation of the route.

Could you also please remove the old sentence below?

sentry/src/routes/channel.rs
```diff
@@ -392,6 +398,112 @@ pub async fn get_accounting_for_channel<C: Locked + 'static>(
...
+        .ok_or(ResponseError::Unauthorized)?
+        .to_owned();
+
+    let spender = auth.uid.to_address();
+
+    let body = hyper::body::to_bytes(req.into_body()).await?;
+    let to_pay = serde_json::from_slice::<ChannelPayRequest>(&body)
+        .map_err(|e| ResponseError::FailedValidation(e.to_string()))?;
+
+    let channel_campaigns = fetch_campaign_ids_for_channel(
```
elpiel, line 435: I've just double checked `fetch_campaign_ids_for_channel`, there is a `expect("TODO")` that should be handled.
Could you please fix it too? Could be in another PR too

sentry/src/routes/channel.rs
```diff
@@ -392,6 +398,112 @@ pub async fn get_accounting_for_channel<C: Locked + 'static>(
...
+    let latest_spendable =
+        fetch_spendable(app.pool.clone(), &spender, &channel_context.context.id())
+            .await
+            .map_err(|err| ResponseError::BadRequest(err.to_string()))?
+            .ok_or(ResponseError::BadRequest("Spendable not found".to_string()))?;
+    let total_deposited = latest_spendable.deposit.total;
+
+    let available_for_return = total_deposited
+        .checked_add(&accounting_earned)
+        .ok_or_else(|| ResponseError::FailedValidation("No more budget remaining".to_string()))?
```
elpiel: should better say that it has overflown when calculating the available amount and similarly - underflow for the subtraction.

sentry/src/routes.rs
```diff
@@ -76,11 +76,17 @@
 //!
 //! Response: [`LastApprovedResponse`][primitives::sentry::LastApprovedResponse]
 //!
-//! - `POST /v5/channel/:id/pay` (auth required)
+//! - [`POST /v5/channel/:id/pay`](channel::channel_payout)(auth required)
+//! Channel Payout with authentication of the spender.
```
elpiel:
```suggestion
//! - [`POST /v5/channel/:id/pay`](channel::channel_payout)(auth required)
//!
//! Channel Payout with authentication of the spender.
//!
```

sentry/src/routes.rs
```diff
@@ -76,11 +76,19 @@
...
+//! all of the earners and updates their balances accordingly. Used when an advertiser/spender wants
+//! to get their remaining funds back.
+//!
+//! Request JSON body: [`ChannelPayRequest`](primitives::sentry::ChannelPayRequest)
+//!
+//! Response: [`SuccessResponse`](primitives::sentry::SuccessResponse)
 //!
-//! Channel Payout with authentication of the spender
 //!
 //! Withdrawals of advertiser funds - re-introduces the PAY event with a separate route.
```
elpiel:
```suggestion
```

sentry/src/routes/campaign.rs
```diff
@@ -115,20 +114,18 @@ pub async fn fetch_campaign_ids_for_channel(
...
+            let skip = i
+                .checked_mul(limit.into())
+                .ok_or(ResponseError::FailedValidation(
+                    "Calculating skip while fetching campaign ids results in an overflow"
+                        .to_string(),
+                ))?;
+            let res = get_campaign_ids_by_channel(pool, &channel_id, limit.into(), skip).await?;
+            other_pages.push(res);
+        }
+        let all_campaigns = other_pages.into_iter().flatten().collect();
```
elpiel: I'm not certain if this will impact performance, since we're now using a for-loop instead of the `try_join_all`, which works with `Future`s.
It should be possible to `return Err(....)` using a `match` and still keep the `try_join_all`

sentry/src/routes/channel.rs
```diff
@@ -392,6 +398,127 @@ pub async fn get_accounting_for_channel<C: Locked + 'static>(
     Ok(success_response(serde_json::to_string(&res)?))
 }
 
+/// `GET /v5/channel/0xXXX.../pay` request
```
```suggestion
/// `POST /v5/channel/0xXXX.../pay` request
```

sentry/src/routes/channel.rs
```diff
@@ -392,6 +398,127 @@ pub async fn get_accounting_for_channel<C: Locked + 'static>(
...
+    let spender = auth.uid.to_address();
+
+    let body = hyper::body::to_bytes(req.into_body()).await?;
+    let to_pay = serde_json::from_slice::<ChannelPayRequest>(&body)
+        .map_err(|e| ResponseError::FailedValidation(e.to_string()))?;
+
+    // Handling the case where a request with an empty body comes through
+    if to_pay.payouts.len() == 0 {
+        return Err(ResponseError::FailedValidation(
+            "Request has an empty body".to_string(),
```
```suggestion
            "Request has empty payouts".to_string(),
```

sentry/src/routes/channel.rs
```diff
@@ -392,6 +398,127 @@ pub async fn get_accounting_for_channel<C: Locked + 'static>(
...
+
+    let total_to_pay = to_pay
+        .payouts
+        .values()
+        .sum::<Option<UnifiedNum>>()
+        .ok_or_else(|| ResponseError::FailedValidation("Payouts amount overflow".to_string()))?;
+
+    if total_to_pay > available_for_return {
+        return Err(ResponseError::FailedValidation(
+            "Total payout amount exceeds the available for return".to_string(),
```
```suggestion
            "The total requested payout amount exceeds the available payout".to_string(),
```

sentry/src/routes/channel.rs
```diff
@@ -392,6 +398,127 @@ pub async fn get_accounting_for_channel<C: Locked + 'static>(
...
+            )
+        })?;
+
+    let total_to_pay = to_pay
+        .payouts
+        .values()
+        .sum::<Option<UnifiedNum>>()
+        .ok_or_else(|| ResponseError::FailedValidation("Payouts amount overflow".to_string()))?;
+
+    if total_to_pay > available_for_return {
```
```suggestion
    if total_to_pay > available_for_payout {
```

sentry/src/routes/channel.rs
```diff
@@ -392,6 +398,127 @@ pub async fn get_accounting_for_channel<C: Locked + 'static>(
...
+    .map(|accounting_spent| accounting_spent.amount)
+    .unwrap_or_default();
+    let latest_spendable =
+        fetch_spendable(app.pool.clone(), &spender, &channel_context.context.id())
+            .await
+            .map_err(|err| ResponseError::BadRequest(err.to_string()))?
+            .ok_or(ResponseError::BadRequest("Spendable not found".to_string()))?;
+    let total_deposited = latest_spendable.deposit.total;
+
+    let available_for_return = total_deposited
```
```suggestion
    let available_for_payout = total_deposited
```

sentry/src/routes/channel.rs
```diff
@@ -846,4 +973,188 @@ mod test {
...
+            channel_context.context.id(),
+            auth.uid.to_address(),
+            deposit.clone(),
+        );
+
+        let res =
+            channel_payout(build_request(&channel_context, UnifiedMap::default()), &app).await;
+        assert!(
+            matches!(res, Err(ResponseError::FailedValidation(..))),
+            "Failed validation because available_for_return is negative"
```
```suggestion
            "Failed validation because available_for_payout is negative"
```

sentry/src/routes/channel.rs
```diff
@@ -846,4 +973,188 @@ mod test {
...
+        payouts.insert(*CREATOR, UnifiedNum::from_u64(1000));
+        app.adapter.client.add_deposit_call(
+            channel_context.context.id(),
+            auth.uid.to_address(),
+            deposit.clone(),
+        );
+        let res = channel_payout(build_request(&channel_context, payouts.clone()), &app).await;
+        assert!(
+            matches!(res, Err(ResponseError::FailedValidation(..))),
+            "Failed validation because total_to_pay > available_for_return"
```
```suggestion
            "Failed validation because total_to_pay > available_for_payout"
```

sentry/src/routes.rs
```diff
@@ -76,11 +76,19 @@
 //!
 //! Response: [`LastApprovedResponse`][primitives::sentry::LastApprovedResponse]
 //!
-//! - `POST /v5/channel/:id/pay` (auth required)
+//! - [`POST /v5/channel/:id/pay`](channel::channel_payout)(auth required)
```
elpiel:
```suggestion
//! - [`POST /v5/channel/:id/pay`](channel::channel_payout) (auth required)
```

sentry/src/routes/channel.rs
```diff
@@ -392,6 +398,127 @@ pub async fn get_accounting_for_channel<C: Locked + 'static>(
...
+            .await
+            .map_err(|err| ResponseError::BadRequest(err.to_string()))?
+            .ok_or(ResponseError::BadRequest("Spendable not found".to_string()))?;
+    let total_deposited = latest_spendable.deposit.total;
+
+    let available_for_return = total_deposited
+        .checked_add(&accounting_earned)
+        .ok_or_else(|| {
+            ResponseError::FailedValidation(
+                "Overflow while calculating available for return".to_string(),
```
```suggestion
                "Overflow while calculating available for payout".to_string(),
```

sentry/src/routes/channel.rs
```diff
@@ -392,6 +398,127 @@ pub async fn get_accounting_for_channel<C: Locked + 'static>(
...
+            )
+        })?
+        .checked_sub(&accounting_spent)
+        .ok_or_else(|| {
+            ResponseError::FailedValidation(
+                "Underflow while calculating available for return".to_string(),
```
```suggestion
                "Underflow while calculating available for payout".to_string(),
```

sentry/src/routes/channel.rs
```diff
@@ -846,4 +973,188 @@ mod test {
...
+            Request::builder()
+                .extension(channel_context.clone())
+                .extension(auth.clone())
+                .body(body)
+                .expect("Should build Request")
+        };
+
+        // Adding a deposit to the DUMMY adapter
+        let deposit = Deposit {
+            total: BigNum::from_str("1000").expect("should convert"), // 1000 DAI
```
elpiel: This is actually not true, this is not 1000 DAI.
You can improve this by using the multiplier from the `app.config`. It is also using `Mocked TOKEN` and not `DAI` particularly (see ganache.toml).

Please, also use non-string variant for `BigNum`s

sentry/src/routes/channel.rs
```diff
@@ -846,4 +973,188 @@ mod test {
...
+            still_on_create2: BigNum::from_str("0").expect("should convert"), // 0 DAI
```
elpiel: You can use `BigNum::from(0_u64)` instead of strings - it's faster.

sentry/src/routes/channel.rs
```diff
@@ -392,6 +398,127 @@ pub async fn get_accounting_for_channel<C: Locked + 'static>(
...
+        Side::Earner,
+    )
+    .await?
+    .map(|accounting_spent| accounting_spent.amount)
+    .unwrap_or_default();
+    let latest_spendable =
+        fetch_spendable(app.pool.clone(), &spender, &channel_context.context.id())
+            .await
+            .map_err(|err| ResponseError::BadRequest(err.to_string()))?
+            .ok_or(ResponseError::BadRequest("Spendable not found".to_string()))?;
```
elpiel: I'm wondering if this is the correct assumption...

It might be better to call `create_or_update_spendable_document` if there is no record in the DB, the same way as in `get_spender_limits`.

Example: Someone that hasn't deposited in the given channel might try to make a payout and he will get a `Spendable not found` error instead of `Total payout amount exceeds the available for payout`.
However, if we create the spendable with `0`, the requester will receive the second error which makes more sense.

What do you think @simzzz ?

sentry/src/routes/channel.rs
```diff
@@ -846,4 +973,156 @@ mod test {
...
+            "Failed validation because request body is empty"
+        );
+        // make a normal request and get accounting to see if its as expected
+
+        let mut payouts = UnifiedMap::default();
+        payouts.insert(*PUBLISHER, UnifiedNum::from_u64(500));
+
+        let res = channel_payout(build_request(&channel_context, payouts.clone()), &app).await;
+        assert!(
+            res.is_ok(),
```
elpiel: It will be more beneficial to assert_eq with Ok(SuccessResponse {}) instead of just `is_ok` this way we will also see an error if for some reason it fails or returns `success: false`.

sentry/src/routes/channel.rs
```diff
@@ -846,4 +973,156 @@ mod test {
...
+            "Request with JSON body with one address and no errors triggered on purpose"
+        );
+        let accounting = get_accounting(
```
elpiel: we should also check the accounting of the spender.
It would be beneficial to also distinguish between the spender & the Publisher accounting in the variable name.

this could also be done in scopes with comments for info
```
// Some info what we expecting and checking for
{
...
}
```

sentry/src/routes/channel.rs
```diff
@@ -846,4 +973,156 @@ mod test {
...
+            matches!(res, Err(ResponseError::FailedValidation(..))),
+            "Failed validation because available_for_return is negative"
+        );
+
+        // make a request where "total_to_pay" will exceed available
+        payouts.insert(*CREATOR, UnifiedNum::from_u64(1000));
+
+        let res = channel_payout(build_request(&channel_context, payouts.clone()), &app).await;
+        assert!(
+            matches!(res, Err(ResponseError::FailedValidation(..))),
```
elpiel: It would be better to have assert_eq and also assert the message returned, as it might fail for other reasons with FailedValidatoin.

sentry/src/routes/channel.rs
```diff
@@ -846,4 +973,156 @@ mod test {
...
+            .increase_by(DUMMY_CAMPAIGN.id, UnifiedNum::from_u64(1000))
+            .await
+            .expect("Should set value in redis");
+
+        let res =
+            channel_payout(build_request(&channel_context, UnifiedMap::default()), &app).await;
+        assert!(
+            matches!(res, Err(ResponseError::FailedValidation(..))),
+            "Failed validation because the campaign has remaining funds"
+        );
```
elpiel: Same goes for this assertion - it should be more specific, otherwise it won't be very helpful for debugging if the assertion fails.

sentry/src/routes/channel.rs
```diff
@@ -846,4 +973,156 @@ mod test {
...
+        update_spendable(app.pool.clone(), &spendable)
+            .await
+            .expect("Should update spendable");
+
+        // Test with no body
+        let res =
+            channel_payout(build_request(&channel_context, UnifiedMap::default()), &app).await;
+        assert!(
+            matches!(res, Err(ResponseError::FailedValidation(..))),
+            "Failed validation because request body is empty"
```
```suggestion
            "Failed validation because payouts are empty"
```

sentry/src/routes/channel.rs
```diff
@@ -846,4 +973,156 @@ mod test {
...
+
+            let body = Body::from(serde_json::to_string(&payouts).expect("Should serialize"));
+
+            Request::builder()
+                .extension(channel_context.clone())
+                .extension(auth.clone())
+                .body(body)
+                .expect("Should build Request")
+        };
+
```
elpiel, line 1021: Could you also perform requests before the rest of the setup is set?

This will make sure that if Accounting or spendable is missing it will still function and return a sane response.

sentry/src/routes/channel.rs
simzzz: Yes, that error message will make more sense although just fetching here would be the simpler operation. If the only way to get this error is when someone hasn't deposited in the given channel we can just change the error message to a one with more clarity, for example, `There is no payout to be returned for this spender on this channel`  @elpiel

sentry/src/routes/campaign.rs
simzzz: With `try_join_all` the closure created by  `map()` expects a Future and `return Err(....)` is a Result. Is there a simple workaround for this?

sentry/src/routes/channel.rs
elpiel: Ok, you can change the error message only to just say "There is no spendable amount for the spender in this Channel."

elpiel (approved):

PR #479, opened 2022-03-16
Tests for GET /v5/channel/list and multichain support

simzzz:


sentry/src/routes/channel.rs
```diff
@@ -642,6 +643,16 @@ mod test {
...
     }
 
+    async fn res_to_channel_list_response(res: Response<Body>) -> ChannelListResponse {
+        let json = hyper::body::to_bytes(res.into_body())
+            .await
+            .expect("Should get json");
+
+        let channel_list_response: ChannelListResponse =
+            serde_json::from_slice(&json).expect("Should get ChannelListResponse");
+        channel_list_response
```
```suggestion
        serde_json::from_slice(&json).expect("Should deserialize ChannelListResponse")
```

sentry/src/routes/channel.rs
```diff
@@ -846,4 +857,86 @@ mod test {
...
+        };
+
+        // Test query page - page 0, page 1
+        let query = ChannelListQuery {
+            page: 0,
+            validator: None,
+        };
+        let res = channel_list(build_request(query), &app).await.expect("should get channels");
+        let channels_list = res_to_channel_list_response(res).await;
+        assert_eq!(channels_list.channels.len(), 2, "There should be 2 channels on the first page");
```
elpiel: We should be asserting the exact channels we expect instead of just `len()` for all tests.

They should be all returned with `ORDER BY created ASC` so assert the order of the channels

sentry/src/routes/channel.rs
```diff
@@ -846,4 +857,86 @@ mod test {
...
+    #[tokio::test]
+    async fn get_channels_list() {
+        let mut app = setup_dummy_app().await;
+        app.config.channels_find_limit = 2;
+
+        let channel = Channel {
+            leader: IDS[&LEADER],
+            follower: IDS[&FOLLOWER],
+            guardian: *GUARDIAN,
+            token: "0x6B175474E89094C44Da98b954EedeAC495271d0F".parse()
```
elpiel: Are these tokens DAI addresses in different chains?

elpiel (changes requested):
Minor things only, good job 😋

sentry/src/routes/channel.rs
simzzz: This one is `chain."Ethereum Mainnet".token.DAI` while the other one is DAI on Goerli. I will clarify this in the expect.

sentry/src/routes/channel.rs
```diff
@@ -846,4 +859,195 @@ mod test {
...
+
+        // Assert that the Ganache chain exist in the configuration
+        let mut config_chain = config
+            .chains
+            .values_mut()
+            .find(|chain_info| chain_info.chain.chain_id == init_chain.chain_id)
+            .expect("Should find Ganache chain in the configuration");
+
+        // override the chain to use the outpace & sweeper addresses that were just deployed
+        config_chain.chain = init_chain.clone();
```
elpiel: You don't need this... GANACHE_1337 load the value from GANACHE_CONFIG, please check source.
If you need to get the token too, you can use the GANACHE_INFO_1337 instead. GANACHE_INFO_1337.tokens["Mocked Token 2"]

sentry/src/routes/channel.rs
```diff
@@ -846,4 +859,195 @@ mod test {
...
+                .tokens
+                .get("Mocked TOKEN")
+                .expect("Should get token")
+                .address,
+            nonce: Nonce::from(987_654_321_u32),
+        };
+        insert_channel(&app.pool, channel)
+            .await
+            .expect("should insert");
+        let channel_other_token = Channel {
```
elpiel, line 1010: This is actually other Chain too. We should put emphasis on the Multi-chain instead of multi-token in this test

sentry/src/routes/channel.rs
```diff
@@ -846,4 +859,195 @@ mod test {
...
+        let channel = Channel {
+            leader: IDS[&LEADER],
+            follower: IDS[&FOLLOWER],
+            guardian: *GUARDIAN,
+            token: config_chain
+                .tokens
+                .get("Mocked TOKEN")
+                .expect("Should get token")
+                .address,
+            nonce: Nonce::from(987_654_321_u32),
```
elpiel, line 1005: Please use different nonce for the channels.

sentry/src/routes/channel.rs
```diff
@@ -846,4 +859,195 @@ mod test {
...
+            let res = channel_list(build_request(query), &app)
+                .await
+                .expect("should get channels");
+            let channels_list = res_to_channel_list_response(res).await;
+            let expected_channels_list = vec![channel, channel_other_token];
+            assert_eq!(
+                channels_list.channels.len(),
+                2,
+                "There should be 1 channel on the second page"
+            );
```
elpiel: These checks are redundant and will not provide enough information on failure.
The assertion with `Vec` will be enough and will provide additional information on missing or additional records that were not expected.

Also use `pretty_assertions`

elpiel (changes requested):
- Fix the conflicting file
- One test is failing because we've added the 0x000 address to config, you can change it to 00..1 until I deploy a new contract on a new Chain Id in ganache.

elpiel (approved):

Issue #471, opened 2022-02-04
Supermarket's `units-for-slot` route integration inside Sentry


elpiel:
## Move Supermarket's `GET /units-for-slot` to sentry

The market will be fully integrated into the platform and the `units-for-slot` route will be moved to Sentry.

`AdSlot` will not be persisted in the database, instead, the whole `AdSlot` will be used in the `adview-manager` to generate the ad view code. 

Initially for an **active status** of the `Campaign` simply use:
- Campaign.active.to > Now
- Is there remaining budget in Campaign (campaign remaining > 0)

Check out the platform issue for more details https://github.com/AmbireTech/adex-platform/issues/625

elpiel:
Here's a non-exhaustive list of the identified changes required to move and alter the `/units-for-slot` route to account for the V5 changes:

- `HTTP` method should be changed to `POST`
- `AdSlot` should be `POST`-ed to the route in order to fetch `AdUnit`s for the particular `Ad type`
- `accepted_referrers`, `categories` & `alexa_rank`, previously taken from the `AdSlotResponse` of the **Market**, should be retrieved from the **Platform** (see https://github.com/AmbireTech/adex-platform/issues/625 for more details)
- Fetch `AdSlot.fallback_unit` from the `Platform` (again, see https://github.com/AmbireTech/adex-platform/issues/625 for more details)
- `depositAsset` query parameters can now be renamed to be an array `depositAsset[]` and even moved to the `POST`-ed data instead of the query.
- Fetch **active** `Campaign`s (**should be discussed what this actually means in V5 for this route**)

elpiel:
# Supermarket statuses to V5
The statuses of campaigns/channels (in **V4**) is computed when they are collected from sentry, before they are put into the cache.
Supermarket statuses, in order of the performed checks, linked to their V5 counter parts (if such exist):

:x: - No counterpart in V5
:question: - uncertain if such counterpart exists in V5

1. Is **Finalized**?
  1.1. Is Channel **expired**? `Utc::now() > channel.valid_until` => Is Campaign active `Utc::now() > campaign.active.to`
  1.2. :x: Is in **withdraw** period? `Utc::now() > channel.spec.withdraw_period_start`
  1.3 Is channel **exhausted**? `total_balances >= channel.deposit_amount` => Has remaining budget (in Campaign)?

2. Is (Channel) **initializing**?
3. Is (Channel) **Unsound**?
  3.1. Is (Channel) offline?
[...]

PR #480, opened 2022-03-18
Issue #471 Route `POST /units-for-slot`

elpiel:
Part of #471 

The initial move of the `GET /units-for-slot` route from the Supermarket to sentry `POST /units-for-slot`

Notable changes to primitives:
- EventType is used for targeting input
- IPFS is used for targeting input instead of string
- Impression & Click Events now have mandatory ad_unit & ad_slot instead of Option
- Campaign's Pricing Bounds now drops the Option and uses a Hashmap instead of struct with separate fields for each EventType


sentry/src/main.rs
```diff
@@ -106,6 +107,14 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
 
     let campaign_remaining = CampaignRemaining::new(redis.clone());
 
+    // todo: Make platform_url configurable! Load from config or pass with env. variable
```
simzzz: We have those values in the config. Can't we just load them from config now?

primitives/src/config.rs
```diff
@@ -37,7 +38,7 @@ impl Default for Environment {
 }
 
 #[derive(Serialize, Deserialize, Debug, Clone)]
-#[serde(rename_all(serialize = "SCREAMING_SNAKE_CASE"))]
+// TODO: use `milliseconds_to_std_duration` for all u32 values with "In milliseconds" in docs.
```
simzzz, line 41: Is the `rename_all` attribute removed on purpose?

sentry/src/db/campaign.rs
```diff
@@ -210,8 +216,51 @@ pub async fn update_campaign(pool: &DbPool, campaign: &Campaign) -> Result<Campa
...
+        }
+        _ => {}
+    };
+
+    // Campaign.creator
+    where_clauses.push("creator = $3".into());
+    params.push(Box::new(creator));
+
+    // Campaign.active.to
+    where_clauses.push("active_to >= $4".into());
```
simzzz: `active_to` is added twice

sentry/src/routes/units_for_slot.rs
```diff
@@ -0,0 +1,422 @@
...
+
+use crate::{
+    db::{
+        accounting::{get_accounting, Side},
+        units_for_slot_get_campaigns, CampaignRemaining, DbPool,
+    },
+    Application, ResponseError,
+};
+
+pub(crate) static CLOUDFLARE_IPCOUNTY_HEADER: Lazy<HeaderName> =
```
simzzz: The variable is `IPCOUNTY`, should be `IPCOUNTRY`

simzzz (changes requested):

primitives/src/config.rs
elpiel: yep, it is. I've decided to drop it since we are no longer bound to the flat structure of the Config

sentry/src/main.rs
elpiel: it's fixed now.

sentry/src/db/campaign.rs
elpiel: fixed.

sentry/src/routes/units_for_slot.rs
elpiel: fixed

primitives/src/campaign.rs
```diff
@@ -217,34 +216,36 @@ impl Campaign {
...
 
     /// Returns the pricing of a given event
-    pub fn pricing(&self, event: &str) -> Option<&Pricing> {
-        self.pricing_bounds
-            .as_ref()
-            .and_then(|bound| bound.get(event))
+    pub fn pricing(&self, event: EventType) -> Option<&Pricing> {
+        self.pricing_bounds.get(&event)
     }
 }
```
elpiel, line 222: This is redundant, so it should be remove soon

primitives/src/campaign.rs
```diff
@@ -253,36 +254,33 @@ mod pricing {
...
-    #[serde(rename_all = "UPPERCASE")]
-    pub struct PricingBounds {
-        #[serde(default, skip_serializing_if = "Option::is_none")]
-        pub impression: Option<Pricing>,
-        #[serde(default, skip_serializing_if = "Option::is_none")]
-        pub click: Option<Pricing>,
-    }
+    #[derive(Default, Debug, PartialEq, Eq, Serialize, Deserialize, Clone)]
+    #[serde(transparent)]
+    pub struct PricingBounds(HashMap<EventType, Pricing>);
```
elpiel, line 259: Pricing bounds is now using the EventType

test_harness/src/lib.rs
```diff
@@ -977,15 +993,25 @@ pub mod run {
...
+        let logger = new_logger(&validator.sentry_logger_prefix);
         let campaign_remaining = CampaignRemaining::new(redis.clone());
 
+        // todo: Make platform_url configurable! Load from config or pass with env. variable
+        let platform_url = "https://todo-local-platform"
+            .parse()
+            .expect("Bad ApiUrl, load from Config?");
+        // todo: Make keep_alive_interval configurable!
+        let platform_api = PlatformApi::new(platform_url, std::time::Duration::from_secs(3))
+            .expect("Should make PlatformApi");
```
simzzz:
```suggestion
       let platform_api = PlatformApi::new(
          config.platform.url.clone(),
          config.platform.keep_alive_interval,
      )
      .expect("Failed to build PlatformApi");
```

We can use the values from config here as well.

simzzz (changes requested):
One more thing I forgot to comment during my previous review. Other than that LGTM.

PR #482, opened 2022-03-28
Issue #336 Multi-chain support - run a second instance of ganache-cli with Chain Id #1

### LINKED ISSUE #336: Multi-chain support (EVM chains)

elpiel:
- scripts - ethereum - Dockerfile - add CHAIN_ID env. var
- scripts - ethereum - add snapshot_contracts_{CHAIN_ID} snapshots
- scropts - ethereum - ganache-cli - use CHAIN_ID env. var.
- docker-compose.harness.yml - run second instance of ganache-cli
- test_harness - Makefile - start second ganache-cli service
- test_harness - add deployment for contracts on second chain
- test_harness - Setup - add logger & logging


test_harness/src/lib.rs
```diff
@@ -28,7 +29,7 @@ pub mod deposits;
 /// NOTE: Current the snapshot and test setup use a single Chain.
 ///
 /// Uses Chain #1337 from the [`GANACHE_CONFIG`] static to init the contracts
-pub static SNAPSHOT_CONTRACTS: Lazy<Contracts> = Lazy::new(|| {
+pub static SNAPSHOT_CONTRACTS_1337: Lazy<Contracts> = Lazy::new(|| {
```
elpiel, line 32: We should think of better way to setup and test when using multi-chain. This is sufficient for the time being.

scripts/ethereum/Dockerfile
```diff
@@ -5,8 +5,14 @@ LABEL maintainer="dev@ambire.com"
 
 WORKDIR /scripts
 
+# The Chain Id that will be passed to `ganache-cli`
+ENV CHAIN_ID=
+# The snapshot contracts directory that will be used
+ARG SNAPSHOT_CONTRACTS=
```
elpiel, line 11: Env. variables cannot be used in `COPY`, this is why we are passing a build argument of the snapshot directory that will be loaded by Ganache-cli

simzzz (approved):

PR #490, opened 2022-04-18
Leader/Follower tick tests + Heartbeat tests

simzzz:
TODO's:
- [x] Fix bug with wiremock
- [x] Add more thorough assertions
- [x] Add better comments
- [x] Add a designated struct to replace the setup_x() functions


primitives/Cargo.toml
```diff
@@ -68,6 +68,7 @@ futures = "0.3"
 async-trait = "0.1"
 # Other
 once_cell = "^1.8"
+wiremock = "0.5.12"
```
elpiel: 1. Should go in dev-dependencies
2. primitives is not the place for mocking server calls.

validator_worker/src/follower.rs
```diff
@@ -333,3 +333,527 @@ async fn on_error<'a, C: Unlocked + 'static>(
...
+    // Initialies a SentryApi instance
+    async fn setup_sentry(server: &MockServer, config: &Config) -> SentryApi<Dummy> {
+        let sentry_url = ApiUrl::from_str(&server.uri()).expect("Should parse");
+
+        let adapter = Adapter::with_unlocked(Dummy::init(Options {
+            dummy_identity: IDS[&LEADER],
+            dummy_auth_tokens: vec![(IDS[&LEADER].to_address(), "AUTH_Leader".into())]
+                .into_iter()
+                .collect(),
+        }));
```
elpiel: You can use the `DUMMY_AUTH` from `test_util`

validator_worker/src/follower.rs
```diff
@@ -333,3 +333,527 @@ async fn on_error<'a, C: Unlocked + 'static>(
...
+        let logger = discard_logger();
+
+        let mut validators: HashMap<ValidatorId, Validator> = HashMap::new();
+        let leader = Validator {
+            url: ApiUrl::from_str(&format!("{}/leader", server.uri())).expect("should be valid"),
+            token: AuthToken::default(),
```
elpiel: And you can use those `DUMMY_AUTH` here as well.
We don't use them but it might be good to take care of these tokens.

validator_worker/src/follower.rs
```diff
@@ -333,3 +333,527 @@ async fn on_error<'a, C: Unlocked + 'static>(
...
+        SentryApi::new(adapter, logger, config.clone(), sentry_url)
+            .expect("Should create instance")
+            .with_propagate(propagate_to)
+            .expect("Should propagate")
+    }
+
+    #[tokio::test]
+    async fn test_follower_tick() {
+        let server_setup = ServerSetup::init(&DUMMY_CAMPAIGN.channel).await;
+        let config = configuration(Environment::Development, None).expect("Should get Config");
```
elpiel: please use `GANACHE_CONFIG` for tests

validator_worker/src/follower.rs
```diff
@@ -333,3 +333,527 @@ async fn on_error<'a, C: Unlocked + 'static>(
...
+        let get_initial_balances = || {
+            let mut balances: Balances<CheckedState> = Balances::new();
+            balances
+                .spend(*ADVERTISER, *PUBLISHER, UnifiedNum::from_u64(1000))
+                .expect("should spend");
+            balances
+                .spend(*ADVERTISER, *PUBLISHER_2, UnifiedNum::from_u64(1000))
+                .expect("should spend");
+            balances
+                .spend(*GUARDIAN, *PUBLISHER, UnifiedNum::from_u64(1000))
```
elpiel: Maybe you can use `CREATOR` instead of `GUARDIAN`, as guardians addresses are primarily used for the Channel.

validator_worker/src/follower.rs
```diff
@@ -333,3 +333,527 @@ async fn on_error<'a, C: Unlocked + 'static>(
...
+
+            let tick_res = tick(
+                &sentry,
+                &channel_context,
+                all_spenders,
+                get_initial_balances(),
+            )
+            .await;
+            assert!(
+                matches!(tick_res, Err(Error::Overflow)),
```
elpiel, line 665: You're right by using `matches`, even though it won't be obvious if a test fails we can't really add `PartialEq` & `Eq` to the Error because of the Adapter's `Error`

validator_worker/src/follower.rs
```diff
@@ -333,3 +333,527 @@ async fn on_error<'a, C: Unlocked + 'static>(
...
+
+        // - Payout mismatch when checking new_state.balances()
+        {
+            let mut new_state_balances = get_initial_balances().into_unchecked();
+            new_state_balances
+                .earners
+                .insert(*GUARDIAN_2, UnifiedNum::from_u64(10000));
+            let new_state: NewState<UncheckedState> = NewState {
+                state_root: String::new(),
+                signature: IDS[&*LEADER].to_checksum(),
```
elpiel, line 678: The ValidatorId is not exactly a signature. Maybe we can use another dummy value instead?

validator_worker/src/follower.rs
```diff
@@ -333,3 +333,527 @@ async fn on_error<'a, C: Unlocked + 'static>(
...
+            };
+            let res = on_new_state(
+                &sentry,
+                &channel_context,
+                get_initial_balances(),
+                new_state,
+                UnifiedNum::from_u64(4000),
+            )
+            .await
+            .expect("Shouldn't return an error");
```
elpiel, line 725: Same for this.

validator_worker/src/follower.rs
```diff
@@ -333,3 +333,527 @@ async fn on_error<'a, C: Unlocked + 'static>(
...
+        {
+            let proposed_balances = get_initial_balances();
+            let state_root = proposed_balances
+                .encode(
+                    channel_context.context.id(),
+                    channel_context.token.precision.get(),
+                )
+                .expect("should encode");
+            let new_state: NewState<UncheckedState> = NewState {
+                state_root: state_root,
```
elpiel: `clippy` will be mad :D, you can use just `stat_root,` in this case

validator_worker/src/follower.rs
```diff
@@ -333,3 +333,527 @@ async fn on_error<'a, C: Unlocked + 'static>(
...
+                UnifiedNum::from_u64(4000),
+            )
+            .await
+            .expect("Shouldn't return an error");
+            assert!(matches!(
+                res,
+                ApproveStateResult::RejectedState {
+                    reason: InvalidNewState::Transition,
+                    ..
+                }
```
elpiel, line 802: Instead of using `matches` for the `*Results` we can `derive(PartialEq, Eq)` and be able to match the exact struct.
Not sure if there is a need for this in these tests, maybe if we add assertions for the propagation results & `state_root` (not mandatory for this PR though).
What do you think?

validator_worker/src/follower.rs
```diff
@@ -333,3 +333,527 @@ async fn on_error<'a, C: Unlocked + 'static>(
...
+            ), "InvalidNewState::Health is the rejection reason when the health penalty is too high");
+        }
+
+        // Case where no NewState is returned
+        {
+            // Setting up the expected response
+            let _mock_guard = server_setup.setup_new_state_response(None).await;
+            let _mock_guard = server_setup.setup_approve_state_response(None).await;
```
elpiel: We have to validate the mock_guard idea in the setup for these tests.
I think shadowing the value like that will drop the guard.

Can we validated that these mock guards work as intended?
And that the setup is not using previously set values/calls?

I think the best way would be to assert exact number of calls (or none) for the particular test case to the wiremock responses.

primitives/src/test_util.rs
```diff
@@ -296,6 +306,190 @@ pub static DUMMY_IPFS: Lazy<[IPFS; 5]> = Lazy::new(|| {
     ]
 });
 
+pub struct ServerSetup {
```
elpiel: This certainly should not be here as it's used in particularly for the validator_worker ticks testing.

validator_worker/src/follower.rs
```diff
@@ -333,3 +333,527 @@ async fn on_error<'a, C: Unlocked + 'static>(
...
+            let _mock_guard = server_setup.setup_new_state_response(Some(new_state)).await;
+
+            let reject_state = RejectState {
+                state_root,
+                signature: IDS[&*FOLLOWER].to_checksum(),
+                timestamp: Some(received),
+                reason: "rejected".to_string(),
+                balances: None,
+            };
+            let _mock_guard = server_setup
```
elpiel: again we have shadowing of the `_mock_guard` variable

validator_worker/src/follower.rs
```diff
@@ -333,3 +333,527 @@ async fn on_error<'a, C: Unlocked + 'static>(
...
+                new_state,
+                UnifiedNum::from_u64(1_000_000),
+            )
+            .await
+            .expect("Shouldn't return an error");
+            assert!(matches!(res, ApproveStateResult::Sent(Some(..))));
+            let propagated_to = match res {
+                ApproveStateResult::Sent(propagated_to) => propagated_to,
+                _ => None, // Shouldn't happen
+            };
```
elpiel: You can just match against:
`ApproveStateResult::Sent(Some(propagated_to))` and `panic!`

elpiel (changes requested):
I haven't reviewed the full PR but check my general remarks in the follower tick

validator_worker/src/follower.rs
simzzz: There isn't a ready-to-use signature in the dummy values. It would get too complicated to generate one for each test case. I think it would be best if we add a few signatures to the dummy values and use them instead

validator_worker/src/follower.rs
simzzz: Deriving `Eq, PartialEq` won't be very straight-forward for `ApproveStateResult` because of the Error types and I think checking just the reason would be sufficient. We can leave it with `matches` for now.

validator_worker/src/follower.rs
simzzz: The `ServerSetup` functions have an `.expect(..)` which asserts the exact number of calls. We can pass it as a function argument for better clarity.

elpiel: Ahhh, yes, this is good then. My only question remaining is if the mock_guard is actually working.
Wiremock says that `When expectations of a scoped [Mock](https://docs.rs/wiremock/latest/wiremock/struct.Mock.html) are not verified, it will trigger a panic` but I can't be certain right now that shadowing the variable like this will actually trigger the drop.
Can we verofy this?

validator_worker/src/follower.rs
elpiel: yes, you're right, Sadly reqwest::Error does not implement either of these traits so it will be difficult to do so.
Let's indeed leave it with a `matches!`.

validator_worker/src/leader.rs
```diff
@@ -127,3 +127,298 @@ async fn on_new_accounting<C: Unlocked + 'static>(
...
+            let proposed_balances = get_initial_balances();
+            let state_root = proposed_balances
+                .encode(
+                    channel_context.context.id(),
+                    channel_context.token.precision.get(),
+                )
+                .expect("should encode");
+            let new_state: NewState<UncheckedState> = NewState {
+                state_root,
+                signature: IDS[&*LEADER].to_checksum(),
```
elpiel, line 387: We can use the Dummy Adapter to sign this NewState, however, I don't think that is really necessary for these tests.

elpiel (approved):
There are a few things we can improve but I think this is ready to be merged.

PR #491, opened 2022-04-26
test_harness - additional multichain tests

elpiel:
Adds additional multi-chain tests:

-  for CAMPAIGN_3 (on Chain # 1) with multiple Events.
- `RejectState` for CAMPAIGN_2 (on Chain # 1337) with multiple Events.
- transitioning from `RejectState` to **unhealthy** `ApproveState` (`is_healthy: false`) for CAMPAIGN_2
- transitioning from **unhealthy** `ApproveState` to **health** for CAMPAIGN_2


adapter/src/ethereum/test_util.rs
```diff
@@ -83,6 +85,15 @@ pub static KEYSTORES: Lazy<HashMap<Address, Options>> = Lazy::new(|| {
             *GUARDIAN_2,
             keystore_options(&format!("{}_keystore.json", *GUARDIAN_2), "ganache6"),
         ),
+        // Address 7
+        // (
+        //     *PUBLISHER_2,
+        //     keystore_options(&format!("{}_keystore.json", *PUBLISHER_2), "ganache7"),
+        // ),
```
elpiel: I've added it as a comment to be easier for addition if the need arises.

docs/config/ganache.toml
```diff
@@ -20,7 +20,9 @@ propagation_timeout = 2000
 
 fetch_timeout = 5000
 all_campaigns_timeout = 5000
-channel_tick_timeout = 5000
+# for test_harness make it larger
+# Default: 5000
+channel_tick_timeout = 8000
```
elpiel, line 25: I still get timeouts in test_harness sometimes, we have to check why this is - performance issue or something wrong with the services

elpiel:
Test is failing as it's comparing the Validator message, which contains a `received` timestamp. This should not be the case and I will make sure to exclude the timestamp from the assertions.

sentry/src/lib.rs
```diff
@@ -510,19 +512,38 @@ pub mod test_util {
...
 
+    /// This guard holds the Redis and Postgres pools taken from their respective Pool of pools.
+    ///
+    /// This ensures that they will not be dropped which will cause tests to fail randomly.
+    pub struct ApplicationGuard {
+        pub app: Application<Dummy>,
+        #[allow(dead_code)]
+        redis_pool: deadpool::managed::Object<crate::db::redis_pool::Manager>,
+        #[allow(dead_code)]
+        db_pool: deadpool::managed::Object<crate::db::tests_postgres::Manager>,
```
elpiel, line 523: Because the `Object` was getting dropped tests failed randomly.
Object hold the Pool of Pools, so we must hold these values until the Application struct get drop itself

simzzz (approved):
The comments made the test logic much easier to understand. I will use a similar approach when writing more complex tests.

PR #492, opened 2022-05-16
Bugfix UnifiedNum math operations

elpiel:


primitives/src/unified_num.rs
```diff
@@ -78,17 +169,51 @@ impl TryFrom<String> for UnifiedNum {
...
+            div_unified_num_to_ratio(self, other).expect("Failed to create ratio for div_floor");
+
+        let whole_number = ratio
+            .checked_div(&Ratio::from_integer(UnifiedNum::MULTIPLIER))
+            .expect("Should divide with Multiplier for div_floor");
+
+        UnifiedNum::from_whole(whole_number.to_integer())
     }
 
+    /// This method returns the inner [`u64`] representation of the [`UnifiedNum`].
```
simzzz: The comment is wrong

sentry/src/routes/units_for_slot_test.rs
```diff
@@ -0,0 +1,761 @@
+use crate::{platform::PlatformApi, test_util::setup_dummy_app};
```
simzzz: I think this file should belong to another PR

simzzz (changes requested):

primitives/src/unified_num.rs
elpiel: Good catch

sentry/src/routes/units_for_slot_test.rs
elpiel: Hmm yeah, I don't know how that got here...

simzzz (approved):

PR #496, opened 2022-06-01
Analytics multichain support

### LINKED ISSUE #494: AIP #61 v5: Analytics multi-chain support
# :heavy_check_mark:  analytics route (`GET /v5/analytics`) #496 
The analytics route should support multi-chains and filter out the record accordingly to the Chain extracted from sentry's authentication.

# :heavy_check_mark:  test_harness #511

The main test cases in `test_harness` are now complete but we are still missing tests regarding the analytics statistics.

Add tests to validate that statistics are recorded correctly for each channel and for all.

simzzz:


primitives/src/analytics.rs
```diff
@@ -115,6 +115,7 @@ pub struct AnalyticsQuery {
     pub hostname: Option<String>,
     pub country: Option<String>,
     pub os_name: Option<OperatingSystem>,
+    pub chains: Option<Vec<ChainId>>,
```
elpiel: just use `Vec` with `#[serde(default)]`

primitives/src/chain.rs
```diff
@@ -21,6 +22,10 @@ impl ChainId {
 
         Self(id)
     }
+
+    pub fn as_u32(self) -> u32 {
```
elpiel: to_u32

elpiel (changes requested):
LGTM, only the OPtion can be made into a Vec with a serde(default) attribute and we can merge this PR :tada:

PR #511, opened 2022-06-17
Included analytics checks in the test_harness module

### LINKED ISSUE #494: AIP #61 v5: Analytics multi-chain support

simzzz:


test_harness/src/lib.rs
```diff
@@ -2308,6 +2775,49 @@ mod tests {
...
+                api_client
+                    .get(endpoint_url)
+                    .bearer_auth(&token)
+                    .send()
+                    .await
+                    .expect("failed to get analytics")
+                    .json::<Vec<FetchedAnalytics>>()
+                    .await
+                    .expect("failed to get json")
+            }
```
elpiel: can we simplify this case by just setting the Auth token for the `/analytics` request instead of calling `for-advertiser`?
It will filter the results accordingly to the token either way.

We can always add additional test for /for-advertiser & for-publisher eventually but that's an overkill for now

test_harness/src/lib.rs
```diff
@@ -2225,6 +2254,444 @@ mod tests {
...
         }
+
+        // Testing analytics statistics cases
+
+        // 1. CHAIN_1 in an unauthenticated /analytics call filters results and correctly returns count/paid
+        {
+            let analytics = get_analytics(
+                &api_client,
+                &leader.sentry_url,
+                &get_analytics_query(Metric::Count, IMPRESSION, vec![ChainId::new(1)]),
```
elpiel: Can you replace these places by using the static values?
E.g. `GANACHE_1337.chain_id` & `GANACHE_1.chain_id`

test_harness/src/lib.rs
```diff
@@ -2225,6 +2254,444 @@ mod tests {
...
+                &get_analytics_query(Metric::Paid, CLICK, vec![ChainId::new(1)]),
+                None,
+            )
+            .await
+            .expect("Should get analytics");
+
+            assert_eq!(analytics.len(), 1, "There should be only one timeframe");
+            assert_eq!(
+                analytics[0].value,
+                FetchedMetric::Paid(UnifiedNum::from_u64(7_000_000)),
```
elpiel: You can make the tests more readable by using `from_whole(0.07)`

elpiel (changes requested):
Great progress .
The test cases look great and according to what we've talked.
We can only tweak a few things (like not calling `analytics/for-advertiser` and cleaning up some code) but I think apart from these small tweaks it's almost ready to be merged.

test_harness/Cargo.toml
```diff
@@ -29,3 +29,4 @@ tokio = { version = "1", features = ["rt-multi-thread", "macros", "time"] }
 # probably needed for Relayer calls
 # wiremock = "0.5"
 pretty_assertions = "^1.0"
+serde_qs = "0.9.2"
```
elpiel, line 32:
```suggestion
serde_qs = "^0.9"
```

You would also need to update cargo.lock though

elpiel (approved):

PR #512, opened 2022-06-17
Channel list filter by chain

### LINKED ISSUE #495: Channel list filtering with chains
Add an array query parameter to `/v5/channel/list` that will filter channels by ChainIds.

We should add the ChainId to the channels table if we need to filter out channels in particular Chain in `GET /v5/channel/list`.

simzzz:


sentry/src/db/channel.rs
```diff
@@ -71,24 +80,36 @@ mod list_channels {
         pool: &DbPool,
         skip: u64,
         limit: u32,
-        validator: Option<ValidatorId>,
+        query: &ChannelListQuery,
```
elpiel: One reason we were not passing the whole ChannelList was because it's related the the API than to the DB and includes the `page` which here is broken down to `skip` & `limit`. so I would suggest (for now) to just pass `&[ChainId]`

sentry/src/db/channel.rs
```diff
@@ -117,25 +138,41 @@ mod list_channels {
...
+        if !query.chains.is_empty() {
+            where_clauses.push(format!(
+                "chain_id IN ({})",
+                query
+                    .chains
+                    .iter()
+                    .map(|id| id.to_u32().to_string())
+                    .collect::<Vec<String>>()
+                    .join(",")
+            ));
```
elpiel: `ChainId` impl From/`ToSql` so you should be able to use it as it is without the `map`

sentry/src/db/spendable.rs
```diff
@@ -134,6 +137,7 @@ mod test {
     #[tokio::test]
     async fn it_inserts_and_fetches_and_updates_spendable() {
         let database = DATABASE_POOL.get().await.expect("Should get a DB pool");
+        let app = setup_dummy_app().await;
 
         setup_test_migrations(database.pool.clone())
             .await
```
elpiel: When using the `setup_dummy_app` it does a lot more work and also runs the `setup_test_migrations`.
I suggest to just use the config directly instead of setting up a whole app: `GANACHE_CONFIG`

sentry/src/db/campaign.rs
```diff
@@ -603,8 +607,14 @@ mod test {
 
         let campaign = DUMMY_CAMPAIGN.clone();
 
+        let channel_chain = app
+            .config
+            .find_chain_of(DUMMY_CAMPAIGN.channel.token)
+            .expect("Channel token should be whitelisted in config!");
+        let channel_context = channel_chain.with_channel(DUMMY_CAMPAIGN.channel);
```
elpiel: I would suggest doing the same thing here as well - use `GANACHE_CONFIG` instead of `setup_dummy_app`

sentry/src/db/accounting.rs
```diff
@@ -481,13 +499,20 @@ mod test {
...
 
         setup_test_migrations(database.pool.clone())
             .await
             .expect("Migrations should succeed");
 
+        let channel_chain = app
+            .config
+            .find_chain_of(DUMMY_CAMPAIGN.channel.token)
+            .expect("Channel token should be whitelisted in config!");
+        let channel_context = channel_chain.with_channel(DUMMY_CAMPAIGN.channel);
```
elpiel: Also here as well - use `GANACHE_CONFIG`

elpiel (changes requested):
Overall pretty good PR.
Apart from the places that introduced `setup_dummy_app` everything else LGTM.
I saw 2 todos & a failing test so I'll wait for those to be done and will check the PR again.

sentry/src/db/channel.rs
simzzz: It needs to be in the format `IN (1, 2, 3)` so I can't use the format macro for Vec<ChainId>. If I do it with params `$1` I get `expected &(dyn ToSql + Sync), found &[ChainId]` so this was the simplest solution I could think of.

elpiel: Fixed, although ugly but at least we can be sure that `u32::MAX` passed to the query won't cause Postgress error since we're using prepared (i.e. `ToSql`)

elpiel (approved):

elpiel:
Closes #495

PR #517, opened 2022-06-24
Removed create2 with sweeper deposits

### LINKED ISSUE #485: Adapter: remove create2 with sweeper deposits

simzzz:
closes #485

### DIFF
--- a/primitives/src/campaign_validator.rs
+++ b/primitives/src/campaign_validator.rs
@@ -24,7 +24,6 @@ pub enum Validation {
     UnlistedValidator,
     UnlistedCreator,
     UnlistedAsset,
-    MinimumDepositNotMet,
     MinimumValidatorFeeNotMet,
     FeeConstraintViolated,
 }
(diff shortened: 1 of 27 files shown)


primitives/src/campaign_validator.rs
```diff
@@ -340,99 +316,99 @@ mod test {
...
-
-            let validation_error = campaign
-                .validate(&config, IDS[&LEADER])
-                .expect_err("Should trigger validation error");
-            assert_eq!(
-                Error::Validation(Validation::FeeConstraintViolated),
-                validation_error,
-            );
-        }
+        // let sum_fees = |validators: &Validators| -> UnifiedNum {
```
elpiel: We should not leave these tests commented out.

primitives/src/config.rs
```diff
@@ -164,8 +164,6 @@ impl ChainInfo {
 /// Precision can differ for the same token from one [`Chain`] to another.
 #[derive(Serialize, Deserialize, Debug, Clone, Eq, PartialEq, Hash)]
 pub struct TokenInfo {
-    #[deprecated = "we no longer need the sweeper contract & create2 addresses for deposits"]
-    pub min_token_units_for_deposit: BigNum,
```
elpiel, line 168: I've made a mistake here. This is not used only for the Create2 deposits but also for the minimum budget of campaigns.
We should keep this configuration and rename it to `min_campaign_budget`.
It should stay in the Token configuration though.

elpiel (changes requested):

elpiel (approved):
Can you also remove the `#![allow(deprecated)]` from the crates?
You can either make it here or in another PR.

PR #520, opened 2022-07-03
Sentry dummy adapter deposits

### LINKED ISSUE #514: Seeding Sentry REST API
Due to the changes made in V5 which relying on `redis` to store the remaining budget for a `Campaign` a postgres seeding of the database is not sufficient for the `sentry` application anymore.

Combined with the fact that we also have a `spendable` amount for each `Channel` that will be regularly update by one of the `Adapter`s (`Dummy` or `Ethereum`) we need to develop a way to fully seed the `sentry` application databases and setup the adapters prerequisites.

In this line of thought, here's what we can do to make the seeding of the development environment of `sentry` usable:

- [x] Create a set of Campaigns (like in test_harness) that will be used to populate the database (AdUnits, Channels, etc) 
   Seeding was introduced in #526 

- [x] A seeder can use a set of Deposits and `Campaign`s (and in turn `Channel`s) to setup the database and adapter in the necessary state
   Seeding with deposits was introduced in #526

- [x] Dummy adapter should be improved upon the current functionality to return a set deposit no matter how often it is called until the value is changed (set a new value through the adapter). The current setup allows only to set a specific call on the Dummy and what value it should return on that specific call only.
   Dummy adapter deposit was introduced in #520 

The only outstanding tasks right now are to:

- [ ] Add the ability to populate events (and in turn analytics and accounting) in those Campaigns programmatically.
- [x] Figure out a way to run the seeding on a `sentry` binary start #559 

## Seed data

We've already introduced 3 campaigns that are available when seeding the sentry application:
`.../rust-validator/sentry$ cargo make run-seed` (runs the `seed` binary)

## Related work

- test_harness - These capabilities are somewhat similar to what the test_harness is doing at the moment (setting Ethereum deposits, making API calls to create campaigns, events, etc.)

- sentry REST API documentation - this setup is also related to the API documentation, since we need to be able to generate examples of the JSON responses, request queries, etc. for the sentry API endpoints.

elpiel:
This branch is based on #518 because of some issues related to clippy warning and seeding the database.

- POST /v5/channel/dummy-deposit - Adds a new channel if it doesn't exist and adds a deposit to the Dummy adapter
- adapter - Dummy - simplifies deposits to be set once (or overwritten later) for the get_deposit call based on Chain
- README - add information about the route for dummy adapter
- routes (REST API docs) - document the new route


adapter/src/dummy.rs
```diff
@@ -190,43 +242,105 @@ impl Unlockable for Dummy {
...
+    /// the set deposit on every call to [`get_deposit()`](crate::client::Locked::get_deposit).
+    #[derive(Debug, Clone, Default)]
+    pub struct Deposits(pub Arc<DashMap<Key, Deposit>>);
+
+    impl Deposits {
+        pub fn new() -> Self {
+            Self::default()
+        }
+
+        /// Get's the set deposit for the give [`ChannelId`] and [`Address`].
```
simzzz: a minor typo, should be

```suggestion
        /// Get's the set deposit for the given [`ChannelId`] and [`Address`].
```

sentry/src/routes/routers.rs
```diff
@@ -228,7 +237,8 @@ pub async fn channels_router<C: Locked + 'static>(
 
         get_accounting_for_channel(req, app).await
     }
-    // POST /v5/channel/:id/pay
+    // POST /v5/channel/:id/dummy-deposit
+    // will allow the calling of the method only if we are using the Dummy adapter!
```
simzzz: This comment is above the `/channel/:id/pay` route

simzzz (changes requested):

sentry/src/routes/routers.rs
elpiel: fixed, good catch!

adapter/src/dummy.rs
elpiel: fixed + changed to Key (the actual used key for the deposits)

simzzz (approved):

PR #523, opened 2022-07-06
Improve CI checks

elpiel:
1. CI workflow - Splits Clippy & rustfmt linting to separate jobs (cargo-make requires nightly for them so run them without CI flow tasks)
2. CI workflow - Adds rustdoc job (for all and no-features) fail on warning
3. Docs workflow - use nightly with `--cfg docsrs`  for features
4. Adapter - regenerate test keystore files using `geth` and `--lightkdf`
5. Run tests in debug mode

This has the added benefit that clippy & rustfmt will not fail the whole process and we can still see if the tests pass.


elpiel:
~~One more improvement I'm looking to do is fetching & building the projects beforehand and sharing the target builds with the rest of the jobs (rustfmt, clippy, docs & cargo-make testing)~~

It turns out that caching is still an issue when it comes to dependencies.
https://github.com/Swatinem/rust-cache offers good starting point but cache size is considerably higher for non-`debug` builds.

PR #544, opened 2022-08-19
GET `/v5/channel/:id/get-leaf` route + tests

simzzz:


simzzz:
TODO:

- [x] Fix te failing test + maybe find a better testing approach
- [x] Documentation for route
- [x] Code examples for GetLeafResponse

sentry/src/routes/channel.rs
```diff
@@ -498,6 +502,98 @@ pub async fn channel_payout<C: Locked + 'static>(
...
+        .balances
+        .spenders
+        .get(&spender)
+        .ok_or_else(|| ResponseError::BadRequest("No balance entry for spender!".to_string()))?;
+    let element = get_balance_leaf(
+        true,
+        &spender,
+        &amount.to_precision(channel_context.token.precision.get()),
+    )
+    .map_err(|err| ResponseError::BadRequest(err.to_string()))?;
```
elpiel: You can remove this `map_err` because ResponseError has an impl From for any Error and will automatically create BadRequest with String error instide.

sentry/src/routes/channel.rs
```diff
@@ -498,6 +502,98 @@ pub async fn channel_payout<C: Locked + 'static>(
...
+
+    let merkle_tree =
+        MerkleTree::new(&[element]).map_err(|err| ResponseError::BadRequest(err.to_string()))?;
+
+    let signable_state_root = get_signable_state_root(&*channel.id(), &merkle_tree.root());
```
elpiel: Use `as_bytes` instead of `&*channel.id(`

sentry/src/routes/channel.rs
```diff
@@ -498,6 +502,98 @@ pub async fn channel_payout<C: Locked + 'static>(
...
+        .ok_or_else(|| ResponseError::BadRequest("No balance entry for spender!".to_string()))?;
+    let element = get_balance_leaf(
+        true,
+        &spender,
+        &amount.to_precision(channel_context.token.precision.get()),
+    )
+    .map_err(|err| ResponseError::BadRequest(err.to_string()))?;
+
+    let merkle_tree =
+        MerkleTree::new(&[element]).map_err(|err| ResponseError::BadRequest(err.to_string()))?;
```
elpiel: same goes for this error mapping

sentry/src/routes/channel.rs
```diff
@@ -1407,4 +1506,119 @@ mod test {
...
+            &MessageTypes::ApproveState(approve_state),
+        )
+        .await
+        .expect("Should insert NewState msg");
+
+        // generate proofs
+        let spender_proof = {
+            let element = get_balance_leaf(
+                true,
+                &*ADVERTISER,
```
elpiel: you probably don\t need `&*`, just `&` is enough

elpiel (changes requested):

elpiel:
Also take into account what we've discussed on slack how to set the earner/spender sides.
`enum LeafFor { Spender, Earner}` and add it to the respective route as an `Extension`

sentry/src/routes.rs
```diff
@@ -237,27 +237,32 @@
 //! ```
 //!
 //!
-//! #### GET `/v5/channel/:id/get-leaf`
-//!
-//! TODO: implement and document as part of issue #382
+//! #### GET `/v5/channel/:id/get-leaf/spender` and GET `/v5/channel/:id/get-leaf/earner`
```
elpiel: This will break the link on top of the docs. I would say to leave it "GET `/v5/channel/:id/get-leaf`" and add the sub-routes in a list below the header `####`, e.g.

```
- GET `/v5/channel/:id/get-leaf/spender/0x....` - for Channel spender leaf
- GET `/v5/channel/:id/get-leaf/earner/0x....` - for Channel earner leaf
```

sentry/src/routes.rs
```diff
@@ -237,27 +237,32 @@
...
 //!
 //! Example Spender:
 //!
-//! `/get-leaf?spender=0x...`
+//! `/get-leaf/spender/0x...`
 //!
 //! Example Earner:
 //!
-//! `/get-leaf?earner=0x....`
+//! `/get-leaf/earner/0x....`
```
elpiel: we can add these examples with actual addresses below in the `Examples` section as e.g. `URI` above `Response`.

sentry/src/routes/channel.rs
```diff
@@ -498,6 +504,68 @@ pub async fn channel_payout<C: Locked + 'static>(
     Ok(Json(SuccessResponse { success: true }))
 }
 
+/// GET `/v5/channel/0xXXX.../get-leaf/spender` request and
+/// GET `/v5/channel/0xXXX.../get-leaf/earner` request
```
elpiel: I would say we should do the same as in the route documentation - "GET `/v5/channel/0xXXX.../get-leaf`" and list the 2 routes for earner & spender below in a list.

sentry/src/routes/channel.rs
```diff
@@ -498,6 +504,68 @@ pub async fn channel_payout<C: Locked + 'static>(
...
+    Path(params): Path<(ChannelId, Address)>,
+) -> Result<Json<GetLeafResponse>, ResponseError> {
+    let channel = channel_context.context;
+
+    let approve_state = match latest_approve_state(&app.pool, &channel).await? {
+        Some(approve_state) => approve_state,
+        None => {
+            return Err(ResponseError::BadRequest(
+                "No ApproveState message for spender".to_string(),
+            ))
```
elpiel: Could you please use 404 in this case. A channel can be new and might not have an ApproveState yet. This shouldn't be a `BadRequest` (400) error.

sentry/src/routes/channel.rs
```diff
@@ -498,6 +504,68 @@ pub async fn channel_payout<C: Locked + 'static>(
...
+        }
+    };
+
+    let state_root = approve_state.msg.state_root.clone();
+
+    let new_state = latest_new_state(&app.pool, &channel, &state_root)
+        .await?
+        .ok_or_else(|| ResponseError::BadRequest("No NewState message for spender".to_string()))?;
```
elpiel, line 531: This should remain the same. If there is an ApproveState, then there must be a NewState that was approved, so 400 is good.

sentry/src/routes/channel.rs
```diff
@@ -498,6 +504,68 @@ pub async fn channel_payout<C: Locked + 'static>(
...
+
+    let addr = params.1;
+
+    let element = match leaf_for {
+        LeafFor::Spender => {
+            let amount = new_state.msg.balances.spenders.get(&addr).ok_or_else(|| {
+                ResponseError::BadRequest("No balance entry for spender!".to_string())
```
elpiel: We should use 404 - Not Found.

sentry/src/routes/channel.rs
```diff
@@ -498,6 +504,68 @@ pub async fn channel_payout<C: Locked + 'static>(
...
+
+            get_balance_leaf(
+                true,
+                &addr,
+                &amount.to_precision(channel_context.token.precision.get()),
+            )?
+        }
+        LeafFor::Earner => {
+            let amount = new_state.msg.balances.earners.get(&addr).ok_or_else(|| {
+                ResponseError::BadRequest("No balance entry for spender!".to_string())
```
elpiel: use 404 here as well

elpiel (changes requested):
Look pretty good. A few polishing touches on Errors remain and a few tests regarding:

- no ApproveState - 404
- no NewState (but has approvestate) - 400
- no spender/earner - you could use the same address and call the function twice

PR #551, opened 2022-09-07
Sentry REST API route - /units-for-slot

### LINKED ISSUE #471: Supermarket's `units-for-slot` route integration inside Sentry

elpiel:


sentry/src/db/campaign.rs
```diff
@@ -216,30 +218,43 @@ pub async fn update_campaign(pool: &DbPool, campaign: &Campaign) -> Result<Campa
...
 }
 
+/// Get Campaigns for GET `/v5/units-for-slot` route.
+///
+/// This method was tested with **5099** deposit assets and it does not reach the postgres query limit.
 pub async fn units_for_slot_get_campaigns(
     pool: &DbPool,
     deposit_assets: Option<&HashSet<Address>>,
-    creator: Address,
+    not_creator: Address,
```
simzzz: I think this naming can be confusing. Maybe leaving it as `creator` or something neutral like `addr`/`publisher_id` would make it clearer.

elpiel: We expect this address to be `!= creator`, not 100% how to make this more clear in the naming, because neutral name might not give the intent of the parameter.
Do you have any other suggestions for the name?

elpiel: Renamed.

simzzz (approved):

PR #558, opened 2022-09-21
Mongodb setup and accounting

elpiel:


primitives/src/sentry.rs
```diff
@@ -335,6 +373,44 @@ pub struct Analytics {
     pub payout_count: u32,
 }
 
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, Eq)]
+#[serde(rename_all = "camelCase")]
+pub struct AnalyticsM {
```
elpiel, line 378: The `M` suffixed structs are used temporary instead of replacing all the usages of the non-M-suffixed structs

primitives/src/sentry.rs
```diff
@@ -1150,6 +1226,38 @@ pub mod campaign_modify {
...
+    }
+
+    impl<Tz> From<DateHour<Tz>> for Bson
+    where
+        Tz: TimeZone,
+    {
+        fn from(date_hour: DateHour<Tz>) -> Self {
+            Bson::DateTime(date_hour.into())
+        }
+    }
```
elpiel, line 1252: There is some sort of an issue with this implementations because using the DateHour instead of Bson::DateTime in query/collections gives an error.

primitives/src/unified_num.rs
```diff
@@ -963,6 +963,20 @@ mod test {
...
+#[cfg(feature = "mongo")]
+mod mongo {
+    use bson::Bson;
+
+    use super::UnifiedNum;
+
+    impl From<UnifiedNum> for Bson {
+        fn from(unified_num: UnifiedNum) -> Self {
+            // TODO: Make it safe to convert to i64!
+            Bson::Int64(i64::try_from(unified_num.0).expect("should convert to i64"))
```
elpiel, line 975: We should try and see if implementing `TryFrom<X> for Bson` will automatically do the conversion in the `doc!` macro and `collection` operations

simzzz (approved):

PR #559, opened 2022-09-29
Seeding sentry REST API for Ethereum and Dummy adapters

### LINKED ISSUE #514: Seeding Sentry REST API

simzzz:


sentry/Cargo.toml
```diff
@@ -15,14 +15,6 @@ default-run = "sentry"
 all-features = true
 rustdoc-args = ["--cfg", "docsrs"]
 
-[[bin]]
-name = "seed"
-test = false # Is tested by default.
-doctest = false # Documentation examples are tested by default.
-bench = false # Is benchmarked by default.
-doc = false # Is documented by default.
-required-features = ["test-util"]
```
elpiel: Add back the `[[bin]]` section and we can merge this.

elpiel (dismissed):
LGTM, just re-add the Cargo.toml section for the binary

elpiel (dismissed):

elpiel (approved):

PR #562, opened 2022-10-13
Adview manager units for slot

simzzz:


adview-manager/src/manager.rs
```diff
@@ -77,6 +80,11 @@ pub struct Options {
     /// default: `false`
     #[serde(default)]
     pub disabled_sticky: bool,
+    /// List of validators to query /units-for-slot from
+    ///
+    /// default: `[]`
+    #[serde(default)]
```
elpiel: I think this field should be mandatory for the operation of the manager.

adview-manager/serve/src/routes.rs
```diff
@@ -97,13 +101,13 @@ pub async fn get_preview_ad(Extension(state): Extension<Arc<State>>) -> Html<Str
         // .and(query_param("depositAsset[]", "0x6B175474E89094C44Da98b954EedeAC495271d03"))
```
elpiel: We should use this line instead but with a fixed name: `depositAssets[]`

adview-manager/src/manager.rs
```diff
@@ -240,40 +248,53 @@ impl Manager {
...
-            }),
+            .validators
+            .clone()
+            .into_iter()
+            .next()
+            .ok_or(Error::NoValidators)?;
+
+        let url = format!(
+            "{}/v5/units-for-slot/{}?{}",
+            first_validator.url, self.options.market_slot, deposit_asset
```
elpiel: After removing the ValidatorDesc, we should use `ApiUrl` instead as it handles the ending `/` when we're joining the rest of the route.

adview-manager/src/manager.rs
```diff
@@ -77,6 +80,11 @@ pub struct Options {
     /// default: `false`
     #[serde(default)]
     pub disabled_sticky: bool,
+    /// List of validators to query /units-for-slot from
+    ///
+    /// default: `[]`
+    #[serde(default)]
+    pub validators: Vec<ValidatorDesc>,
```
elpiel: It does not seem that we use any other part from ValidatorDesc except the Url.

If this is the case, we don't need `ValidatorDesc` but also we should remove it, as it's part of the Campaign itself and can't be used as a common struct (fee, fee_addr, etc).

adview-manager/src/manager.rs
```diff
@@ -240,40 +248,53 @@ impl Manager {
...
+        let mut first_res: Response = serde_json::from_str(&json_res).expect("Should convert");
+
+        for validator in self.options.validators.iter().skip(1) {
+            let url = format!(
+                "{}/v5/units-for-slot/{}?{}",
+                validator.url, self.options.market_slot, deposit_asset
+            );
+            let new_res: Response = self.client.get(url).send().await?.json().await?;
+            for campaign in new_res.campaigns {
+                if !first_res.campaigns.contains(&campaign) {
```
elpiel: We should base this comparison using the `CampaignId`. Remember that we want to avoid discrepancies in `Campaign`s. I.e. a bad validator might say the budget is 2000 tokens instead of 100 and the Hash of those two Campaigns will be different, so `contains()` will return false.

adview-manager/src/manager.rs
```diff
@@ -240,40 +248,53 @@ impl Manager {
...
+            .ok_or(Error::NoValidators)?;
+
+        let url = format!(
+            "{}/v5/units-for-slot/{}?{}",
+            first_validator.url, self.options.market_slot, deposit_asset
+        );
+
+        // Ordering of the campaigns matters so we will just push them to the first result
+        // We reuse `targeting_input_base`, `accepted_referrers` and `fallback_unit`
+        let json_res: String = self.client.get(url).send().await?.text().await?;
```
elpiel: you should use `.json()` instead of `text()`.

adview-manager/src/manager.rs
```diff
@@ -240,40 +248,53 @@ impl Manager {
         }
     }
 
-    pub async fn get_market_demand_resp(&self) -> Result<Response, Error> {
-        let pub_prefix = self.options.publisher_addr.to_hex();
-
+    // Test with different units with price
+    // Test if first campaign is not overwritten
+    pub async fn get_units_for_slot_resp(&self) -> Result<Response, Error> {
         let deposit_asset = self
```
elpiel: I think it should be plural to reflect what it does - filtering by deposit assets.

adview-manager/src/manager.rs
```diff
@@ -418,3 +439,193 @@ impl Manager {
...
+            target_url: unit_0.target_url,
+        };
+
+        let unit_1 = DUMMY_AD_UNITS[1].clone();
+        let modified_ad_unit = AdUnit {
+            ipfs: unit_1.ipfs,
+            media_url: unit_1.media_url,
+            media_mime: unit_1.media_mime,
+            target_url: unit_1.target_url,
+        };
```
elpiel: There's an Into impl for the AdUnit of units-for-slot but it uses a reference to `AdUnit`. You can use that instead.

adview-manager/src/manager.rs
```diff
@@ -418,3 +439,193 @@ impl Manager {
...
+
+        let campaign_0 = Campaign {
+            campaign: CAMPAIGNS[0].clone().context,
```
elpiel: A nit-pick - `context.clone()`?

adview-manager/src/manager.rs
```diff
@@ -418,3 +439,193 @@ impl Manager {
...
+        let publisher_addr = "0x0000000000000000626f62627973686d75726461"
+            .parse()
+            .unwrap();
+        let mut validator_1 = DUMMY_VALIDATOR_LEADER.clone();
+        validator_1.url = format!("{}/1", server.uri());
+        let mut validator_2 = DUMMY_VALIDATOR_FOLLOWER.clone();
+        validator_2.url = format!("{}/2", server.uri());
+        let mut validator_3 = DUMMY_VALIDATOR_LEADER.clone();
+        validator_3.id = IDS[&LEADER_2];
+        validator_3.url = format!("{}/3", server.uri());
```
elpiel: I think we can benefit from being more explicit, e.g. prefix it with `validator-1` for example

adview-manager/src/manager.rs
```diff
@@ -418,3 +439,193 @@ impl Manager {
...
+        let options = Options {
+            market_url,
+            market_slot: DUMMY_IPFS[0],
+            publisher_addr,
+            // All passed tokens must be of the same price and decimals, so that the amounts can be accurately compared
```
elpiel, line 589: we can remove this comment from here. I'm actually not 100% if that's true anymore, as well.

adview-manager/src/manager.rs
```diff
@@ -240,40 +249,60 @@ impl Manager {
...
+                    self.options.market_slot, deposit_assets
+                ))
+                .map_err(|_| Error::InvalidValidatorUrl)?;
+            let new_res: Response = self.client.get(url.as_str()).send().await?.json().await?;
+            for response_campaign in new_res.campaigns {
+                if !first_res
+                    .campaigns
+                    .iter()
+                    .map(|rc| rc.campaign.id)
+                    .any(|x| x == response_campaign.campaign.id)
```
elpiel: You can directly use `any` instead of mapping the ID first

adview-manager/src/manager.rs
```diff
@@ -240,40 +249,60 @@ impl Manager {
...
-            StatusCode::OK => Ok(market_response.json().await?),
-            _ => Err(Error::Market {
-                status: market_response.status(),
-                url,
-            }),
+            .map_err(|_| Error::InvalidValidatorUrl)?;
+        // Ordering of the campaigns matters so we will just push them to the first result
+        // We reuse `targeting_input_base`, `accepted_referrers` and `fallback_unit`
+        let mut first_res: Response = self.client.get(url.as_str()).send().await?.json().await?;
+        // let mut first_res: Response = serde_json::from_str(&json_res).expect("Should convert");
```
elpiel: We can now remove this commnet

adview-manager/src/manager.rs
```diff
@@ -240,40 +249,60 @@ impl Manager {
...
-        // ApiUrl handles endpoint path (with or without `/`)
-        let url = self
+        let first_validator = self
             .options
-            .market_url
+            .validators
+            .clone()
+            .into_iter()
+            .next()
+            .ok_or(Error::NoValidators)?;
```
elpiel: maybe it would be easier to just use `get(0).clone().ok_or(...)`?

adview-manager/src/manager.rs
```diff
@@ -418,3 +447,176 @@ impl Manager {
...
+        Mock::given(method("GET"))
+            .and(path(format!("validator-3/v5/units-for-slot/{}", slot,)))
+            .respond_with(ResponseTemplate::new(200).set_body_json(&response_3))
+            .mount(&server)
+            .await;
+
+        // 2. Set up a manager
+        let market_url = server.uri().parse().unwrap();
+        let whitelisted_tokens = DEFAULT_TOKENS.clone();
+        let publisher_addr = "0x0000000000000000626f62627973686d75726461"
```
elpiel: Should we just use the `PUBLISHER` instead?

adview-manager/serve/src/routes.rs
```diff
@@ -84,26 +89,26 @@ pub async fn get_preview_ad(Extension(state): Extension<Arc<State>>) -> Html<Str
         campaigns: vec![],
     };
 
-    // Mock the `get_market_demand_resp` call
+    // Mock the `get_units_for_slot_resp` call
     let mock_call = Mock::given(method("GET"))
-        // &depositAsset={}&depositAsset={}
+        // &depositAssets[]={}&depositAssets[]={}
```
elpiel: we can remove this comment all together

adview-manager/serve/src/routes.rs
```diff
@@ -84,26 +89,26 @@ pub async fn get_preview_ad(Extension(state): Extension<Arc<State>>) -> Html<Str
...
-        // pubPrefix=HEX&depositAsset=0xASSET1&depositAsset=0xASSET2
+        // pubPrefix=HEX&depositAssets[]=0xASSET1&depositAssets[]=0xASSET2
         .and(query_param("pubPrefix", pub_prefix))
         .and(query_param(
-            "depositAsset",
+            "depositAssets[]",
             "0x6B175474E89094C44Da98b954EedeAC495271d0F",
         ))
-        // .and(query_param("depositAsset[]", "0x6B175474E89094C44Da98b954EedeAC495271d03"))
+        // .and(query_param("depositAssets[]", "0x6B175474E89094C44Da98b954EedeAC495271d03"))
```
elpiel: and this comment too. It was added before the units-for-slot was finished and that's why it was commented out. Now we have it above.

elpiel (approved):

PR #566, opened 2022-10-26
Infinity Behaviour + AdView Manager Tests

simzzz:
2 TODO's in `get_sticky_ad_unit()`

### DIFF
--- a/adview-manager/src/helpers.rs
+++ b/adview-manager/src/helpers.rs
@@ -163,7 +163,7 @@ pub fn get_unit_html_with_events(
             .iter()
             .map(|validator| {
                 let fetch_url = format!(
-                    "{}/campaign/{}/events?pubAddr={}",
+                    "{}/v5/campaign/{}/events?pubAddr={}",
                     validator.url, campaign_id, options.publisher_addr
                 );
 
@@ -201,12 +201,13 @@ pub fn get_unit_html_with_events(
 #[cfg(test)]
 mod test {
     use super::*;
-    use crate::manager::DEFAULT_TOKENS;
     use adex_primitives::{
+        config::GANACHE_CONFIG,
         test_util::{DUMMY_CAMPAIGN, DUMMY_IPFS, PUBLISHER},
         util::ApiUrl,
     };
     use scraper::{Html, Selector};
+    use std::collections::HashSet;
 
     fn get_ad_unit(media_mime: &str) -> AdUnit {
         AdUnit {
@@ -354,7 +355,11 @@ mod test {
 
     #[test]
     fn getting_unit_html_with_events() {
-        let whitelisted_tokens = DEFAULT_TOKENS.clone();
+        let whitelisted_tokens = GANACHE_CONFIG
+            .chains
+            .values()
+            .flat_map(|chain| chain.tokens.values().map(|token| token.address))
+            .collect::<HashSet<_>>();
 
         let market_url = ApiUrl::parse("https://market.adex.network").expect("should parse");
         let validator_1_url = ApiUrl::parse("https://tom.adex.network").expect("should parse");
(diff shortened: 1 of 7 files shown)


simzzz:
skipped testing `get_next_ad_unit()` due to the complexity of testing all the cases, will test it as well if necessary

adview-manager/src/helpers.rs
```diff
@@ -231,6 +236,185 @@ mod test {
...
+            assert_eq!(
+                Some("https://ambire.com?utm_source=AdEx+(https://adsense.google.com)"),
+                anchor.value().attr("href")
+            );
+        }
+    }
+
+    #[test]
+    fn getting_unit_html_with_events() {
+        let whitelisted_tokens = DEFAULT_TOKENS.clone();
```
elpiel: For the whitelisted tokens we can use `GANACHE_CONFIG` instead:

```
let whitelisted_tokens = config
        .chains
        .values()
        .flat_map(|chain| chain.tokens.values().map(|token| token.address))
        .collect::<Vec<_>>()
```

We can add a separate static in a different branch so I can use it too.

adview-manager/src/helpers.rs
```diff
@@ -231,6 +236,185 @@ mod test {
...
+                .expect("There should be a second anchor");
+
+            let video_selector = Selector::parse("div>a>video").unwrap();
+            let video = fragment
+                .select(&video_selector)
+                .next()
+                .expect("There should be a video");
+
+            // TODO: If campaign.validators doesn't guarantee order this might fail and become untestable
+            let expected_onclick: &str = &format!("var fetchOpts = {{ method: 'POST', headers: {{ 'content-type': 'application/json' }}, body: {{'events':[{{'type':'CLICK','publisher':'{}','adUnit':'{}','adSlot':'{}','referrer':'document.referrer'}}]}} }}; fetch('{}/campaign/{}/events?pubAddr={}', fetchOpts); fetch('{}/campaign/{}/events?pubAddr={}', fetchOpts)", options.publisher_addr, ad_unit.ipfs, options.market_slot, validators.iter().nth(0).unwrap().url, campaign_id, options.publisher_addr, validators.iter().nth(1).unwrap().url, campaign_id, options.publisher_addr);
```
elpiel: You can just use `validators[0].url` and `validators[1].url`.
Index is impled for Validators so this is possible

adview-manager/src/helpers.rs
```diff
@@ -231,6 +236,185 @@ mod test {
...
+            assert_eq!(Some(expected_onclick), anchor.value().attr("onclick"));
+
+            let expected_onloadeddata: &str = &format!("setTimeout(function() {{ var fetchOpts = {{ method: 'POST', headers: {{ 'content-type': 'application/json' }}, body: {{'events':[{{'type':'IMPRESSION','publisher':'{}','adUnit':'{}','adSlot':'{}','referrer':'document.referrer'}}]}} }}; fetch('{}/campaign/{}/events?pubAddr={}', fetchOpts); fetch('{}/campaign/{}/events?pubAddr={}', fetchOpts) }}, 8000)", options.publisher_addr, ad_unit.ipfs, options.market_slot, validators.iter().nth(0).unwrap().url, campaign_id, options.publisher_addr, validators.iter().nth(1).unwrap().url, campaign_id, options.publisher_addr);
```
elpiel: same goes for this one

primitives/src/targeting/input.rs
```diff
@@ -127,9 +127,11 @@ impl GetField for AdView {
 
     fn get(&self, field: &Self::Field) -> Self::Output {
         match field {
-            field::AdView::SecondsSinceCampaignImpression => {
-                Value::Number(self.seconds_since_campaign_impression.into())
-            }
+            field::AdView::SecondsSinceCampaignImpression => Value::Number(
```
elpiel, line 132:
```suggestion
            // We could use `Option`, however, `try_get` will return `UnknownVariable`
            // this is why we use `u64::MAX` when returning the value of the field.
            field::AdView::SecondsSinceCampaignImpression => Value::Number(
```

adview-manager/src/helpers.rs
```diff
@@ -231,6 +236,185 @@ mod test {
...
+                .next()
+                .expect("There should be a second anchor");
+
+            let video_selector = Selector::parse("div>a>video").unwrap();
+            let video = fragment
+                .select(&video_selector)
+                .next()
+                .expect("There should be a video");
+
+            // TODO: If campaign.validators doesn't guarantee order this might fail and become untestable
```
elpiel: This is correct, `campaign.validators` does **not** guarantees the order will be `leader, follower`!

adview-manager/src/manager.rs
```diff
@@ -607,4 +604,86 @@ mod test {
...
+            .await;
+
+        assert!(res.is_some());
+
+        // TODO: Here we modify the campaign manually, verify that such a scenario is possible
+        let campaign = Campaign {
+            campaign: DUMMY_CAMPAIGN.clone(),
+            units_with_price: vec![UnitsWithPrice {
+                unit: AdUnit::from(&DUMMY_AD_UNITS[1]),
+                price: UnifiedNum::from_u64(10_000),
```
elpiel: Please use `from_whole` with a float.

adview-manager/src/manager.rs
```diff
@@ -607,4 +604,86 @@ mod test {
...
+            }],
+        };
+
+        // TODO: This panics, verify that this is the correct behaviour
```
elpiel: Since we've validated that it panics, then I would say to make it return `None` instead.

I will check with Ivo too but this makes more sense.

adview-manager/src/manager.rs
```diff
@@ -607,4 +604,86 @@ mod test {
...
+        manager.history = history;
+
+        let campaign = Campaign {
+            campaign: DUMMY_CAMPAIGN.clone(),
+            units_with_price: vec![UnitsWithPrice {
+                unit: AdUnit::from(&DUMMY_AD_UNITS[0]),
+                price: UnifiedNum::from_u64(10_000),
+            }],
+        };
+        let res = manager
```
elpiel, line 670: I'm not 100% sure if this panics whether the test will pass. This is something we have to be careful with.
This might be panicking instead of the second call.

adview-manager/src/helpers.rs
elpiel: - [ ] I found a bug in the old code, it should be `/v5/campaign/...`
please fix this, as well as, the tests

elpiel (approved):

PR #572, opened 2022-11-17
Cleaning up remaining todo's in the code

### LINKED ISSUE #570: Remaining TODO's in the code
### Adapter

- [x] `ethereum/client.rs` - We return an `EwtSigningError` when the wallet sign fails in the `sign()` function even though we do not sign an Ethereum Web Token but an Outpace state_root
- [x] `ethereum/ewt.rs` - Determine if it's necessary to replace the error in the `base64_decode()` function for the `Payload` struct with our own instead of `EwtVerifyError`

### Ad-view Manager

- [x] - `serve/src/routes.rs` -  Check if the `navigator_language` value in the options is correct
- [x] - `helpers.rs` - Check if the quotes in the `fetch_opts` variable are correct
_Done in #564_

- [x] - `manager.rs` - Check what should happen if no validators are found in `get_next_adunit()` for the campaign
_Checked this while working on #566 , there should always be a validator_

- [x] - `manager.rs` - Check if there is a scenario in production where the `units_with_price` for a campaign can be modified

### Config

~~`ganache.toml` - Check the POST /validator-message route performance, more than 1 second for timeout is a lot!~~

### Primitives

~~`analytics/query.rs` - Make the `start` field in the `Time` struct either a timestamp or `DateTime` string when serializing/deserializing~~
_Mentioned in #493, currently low priority_
- [x] - `targeting/input/field.rs` - `Balances` enum should drop `PublisherEarnedFromCampaign` value 
- [x] - `targeting/eval.rs` - Check the issue linked above the `Function` enum
- [x] - `lib.rs` - Remove `allow(unstable_name_collisions)` once num::Integer::div_floor() is fixed
- [x] - `unified_num.rs` - Check math and write tests for the `mod_floor()`, `gcd()`, `lcm()`, `divides()`, `is_multiple_of(), and `div_rem()` functions
_.divides() is deprecated, should we keep it implemented?_
- [x] - `cargo.toml` - Change ChannelId Serialize implementation in order to remove `serde-hex`
_How should we serialize it instead?_

### Sentry

- [x] - `db/analytics.rs` - throw a few more analytics in specific DateHours w/ unique Query and check if it filters the analytics correctly in the `test_get_analytics_in_december()` test
- [x] - `db/campaign.rs` - Check and update the statement in `fetch_campaign()`

~~#571~~
_Needs more research, exporting to a separate issue_

~~`db.rs` - Fix error mapping in the `create()` function of the `Manager`~~

### Validator Worker
- [x] - `error.rs` - Add channel, validatorId etc. for the `ChannelNotIntendedForUs` Error variant in the enum
~~`follower.rs` - Check if we should use the health value when displaying `InvalidNewState`~~
- [x]  - `follower.rs` - Context for All spender sum Error when overflow occurs in the follower tick
~~`follower.rs` - Should we show the Payout Mismatch between Spent & Earned in `on_new_state()`~~
- [x] - `follower.rs` - Add context for `proposed_balances.check()` overflow error
- [x] - `follower.rs` - Add context for Transition error
- [x] - `follower.rs` - Add context for error in Spenders and Earners transition
- [x] - `lib.rs` - Double check `get_state_root_hash_returns_correct_hash` test
_How should the encoded value be checked?_

simzzz:
closes #570

### DIFF
--- a/primitives/src/unified_num.rs
+++ b/primitives/src/unified_num.rs
@@ -309,27 +309,22 @@ impl Integer for UnifiedNum {
         UnifiedNum::div_floor(self, other)
     }
 
-    // TODO: Check math and write tests
     fn mod_floor(&self, other: &Self) -> Self {
         self.0.mod_floor(&other.0).into()
     }
 
-    // TODO: Check math and write tests
     fn gcd(&self, other: &Self) -> Self {
         self.0.gcd(&other.0).into()
     }
 
-    // TODO: Check math and write tests
     fn lcm(&self, other: &Self) -> Self {
         self.0.lcm(&other.0).into()
     }
 
-    // TODO: Check math and write tests
     fn divides(&self, other: &Self) -> bool {
         self.0.divides(&other.0)
     }
 
-    // TODO: Check math and write tests
     fn is_multiple_of(&self, other: &Self) -> bool {
         self.0.is_multiple_of(&other.0)
     }
@@ -342,7 +337,6 @@ impl Integer for UnifiedNum {
         !self.is_even()
     }
 
-    // TODO: Check math and write tests
     fn div_rem(&self, other: &Self) -> (Self, Self) {
         let (quotient, remainder) = self.0.div_rem(&other.0);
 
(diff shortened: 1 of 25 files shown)


primitives/src/unified_num.rs
```diff
@@ -961,6 +955,69 @@ mod test {
             );
         }
     }
+
+    #[test]
+    fn test_unified_num_mod_floor_gcd_lcm_divides_is_multiple_of_div_rem() {
```
elpiel, line 960: I'm uncertain about these tests. You're actually checking `0.00000008` value in the first test for example, not integers so the math is different for floating numbers.
Another things I realised is that `num::Integer` allows these methods and it's not implemented for floating numbers, so maybe we don't need to include `Integer` impl for `UnitfiedNum`

elpiel (changes requested):

